PR #2, opened 2018-07-16
Functions to freeze bundles

alec-deason:
This adds freezing of bundle data. Also exploring how to test and build APIs around the execution context.


alec-deason:
I'm withdrawing this because I've discovered that my assumptions about where canonical bundle data is stored are wrong and this method for freezing (which I borrowed from dismodmr) will not actually work for dismodat.

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
+from contextlib import contextmanager
```
nrhawkins: We have been adding a docstring at the top of every module.

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,238 @@
+import logging
```
nrhawkins: We have been adding a docstring at the top of every module.

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
...
+            raise ImportError(f"Required package db_tools not found")
+
+    ezfuncs = DummyEZFuncs()
+
+
+@contextmanager
+def cursor(execution_context=None, database=None):
+    """A context manager which exposes a database cursor connected to the database specified by
+    the execution_context. The cursor will be closed when the manager exits and if it exits without
+    raising an exception the connection will also be commited.
```
nrhawkins: commited -> committed

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
+from contextlib import contextmanager
+
+try:
```
nrhawkins: I don't think I have any try's in my imports, so this is making me wonder if/when I should.

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
...
+        def __getattr__(self, name):
+            raise ImportError(f"Required package db_tools not found")
+
+    ezfuncs = DummyEZFuncs()
+
+
+@contextmanager
+def cursor(execution_context=None, database=None):
+    """A context manager which exposes a database cursor connected to the database specified by
+    the execution_context. The cursor will be closed when the manager exits and if it exits without
```
nrhawkins: The first sentence is making me wonder if database is needed, how it's used.

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
...
+    raising an exception the connection will also be commited.
+    """
+
```
nrhawkins: I see, the below sets database, if context_manager is given.

src/cascade/core/db.py
```diff
@@ -0,0 +1,55 @@
...
+    if execution_context is None:
+        if database is None:
+            raise ValueError("Must supply either execution_context or database")
+    else:
+        if database is not None:
+            raise ValueError("Must not supply both execution_context and database")
+        database = execution_context.parameters.database
+
```
nrhawkins: If context_manager wasn't supplied, and database was, database will be used here.

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,238 @@
...
+        covariate_rows = list(c)
+        CODELOG.debug(
+            f"Downloaded {len(covariate_rows)} lines of study covariates for bundle_id {bundle_id} from '{database}'"
+        )
+
+    return covariate_rows
+
+
+def _upload_bundle_data_to_tier_3(cursor, model_version_id, bundle_data):
+    """Updloads bundle data to tier 3 attached to the current model_version_id.
```
nrhawkins: Updloads -> uploads

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,238 @@
...
+    )
+    """
+
+    cursor.executemany(insert_query, bundle_data)
+
+    CODELOG.debug(f"uploaded {len(bundle_data)} lines of bundle data")
+
+
+def _upload_study_covariates_to_tier_3(cursor, model_version_id, covariate_data):
+    """Updloads study covariate mappings to tier 3 attached to the current model_version_id.
```
nrhawkins: Updloads -> Uploads

src/cascade/core/db.py
alec-deason: Ok. I'll do better documentation :.(

src/cascade/core/db.py
alec-deason: In general you shouldn't. I'm actually not 100% it's the best way of doing it here though this is a relatively common pattern. What I'm trying to do is make the dependency on ezfuncs optional. This allows test code or other systems that want to use this module to import it without actually having to install any of the central comp stuff. But if you then go and do something that actually does require ezfuncs you'll get an immediate error explaining that it isn't installed.

The other common way of doing this is to have the import inside the functions where it's actually used. But that often leads to having several copies of the same import scattered throughout the module, which is ugly and confusing.

src/cascade/core/db.py
alec-deason: Yeah, I agree.

src/cascade/core/db.py
alec-deason: Yep. This is because in the majority of cases we want the dismod_at database but occasionally we will need to access other databases, like the central epi database. I've had some discussions with CC since writing this which make it seem like we may eventually stop needing to access the epi database directly but for now it's still required.

src/cascade/input_data/db/bundle.py
alec-deason: I think that's actually an independent typo from the one above. I don't believe I copied and pasted this.

PR #3, opened 2018-07-20
Feature/dismodat 256 download bundle

alec-deason:
Functions to download raw bundle and study covariate data.


src/cascade/core/db.py
```diff
@@ -0,0 +1,66 @@
...
+            raise ImportError(f"Required package db_tools not found")
+
+    ezfuncs = DummyEZFuncs()
+
+
+@contextmanager
+def cursor(execution_context=None, database=None):
+    """A context manager which exposes a database cursor connected to the database specified by
+    the execution_context. The cursor will be closed when the manager exits and if it exits without
+    raising an exception the connection will also be commited.
```
nrhawkins: commited -> committed

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+
+def _bundle_is_frozen(execution_context):
+    """Checks if data for the current model_version_id already exists in tier 3.
+    """
+
+    model_version_id = execution_context.parameters.model_version_id
+    query = f"""
+    select exists(
+             select * from epi.t3_model_version_dismod
+             where model_version_id = {model_version_id}
```
nrhawkins: Is one of these methods preferred?  
i.e. where model_version_id = {model_version_id}   vs.  where model_version_id = %(model_version_id)s

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+
+        if len(bundle_ids) > 1:
+            raise ValueError(f"Multiple bundle_ids associated with model_version_id {model_version_id}")
+
+        return bundle_ids[0][0]
+
+
+def _get_bundle_data(execution_context, bundle_id, tier=3):
+    """Downloads the tier 2 data for the bundle associated with the current model_version_id.
+    """
```
nrhawkins: should this be "tier 2 or tier 3 data" ?

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+         """
+    with connection(database=database) as c:
+        bundle_data = pd.read_sql(query, c, params={"bundle_id": bundle_id})
+        CODELOG.debug(f"Downloaded {len(bundle_data)} lines of bundle_id {bundle_id} from '{database}'")
+
+    return bundle_data
+
+
+def _get_study_covariates(execution_context, bundle_id, tier=3):
+    """Downloads the tier 2 study covariate mappings for the bundle associated with the current model_version_id.
```
nrhawkins: should this be "tier 2 or tier 3 study covariate mappings"?

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+    )
+    """
+
+    cursor.executemany(insert_query, bundle_data.values())
+
+    CODELOG.debug(f"uploaded {len(bundle_data)} lines of bundle data")
+
+
+def _upload_study_covariates_to_tier_3(cursor, model_version_id, covariate_data):
+    """Updloads study covariate mappings to tier 3 attached to the current model_version_id.
```
nrhawkins: Updloads -> Uploads

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+    cols = ["measure", "mean", "sex", "standard_error", "age_start", "age_end", "year_start", "year_end"]
+
+    return data[cols]
+
+
+def _covariate_ids_to_names(execution_context, study_covariates):
+    """Convert study_covariate_ids to canonical study covariate names
+    """
+    query = """
+    select study_covariate_id, study_covariate
```
nrhawkins: Should we always capitalize these; i.e. SELECT?

src/cascade/input_data/db/bundle.py
```diff
@@ -0,0 +1,327 @@
...
+
+    Returns:
+        A tuple of (bundle data, study covariate labels) where the bundle data is a pd.DataFrame and the labels are a
+        pd.Series with an index aligned with the bundle data
+    """
+    if bundle_id is None:
+        bundle_id = _get_bundle_id(execution_context)
+
+    bundle = _get_bundle_data(execution_context, bundle_id, tier=tier)
+    bundle = _normalize_bundle_data(bundle)
```
nrhawkins: It has been mentioned to me in a PR that I shouldn't re-use a variable name, pointing to Code Complete for the justification.  

Any thoughts on that?

src/cascade/input_data/db/bundle.py
alec-deason: Yes. This is a mistake. It should be %(model_version_id)s. The difference is in when the substitution happens. With the f-string version the substitution is done by python without any awareness of SQL. In the other version the substitution is done by the SQL client itself which allows it to protect itself from values which might have bad escaping in them or other things that could unexpectedly render the SQL statement incorrect.

In applications which are more exposed to untrusted users than us, this is a serious concern because the f-string version might allow someone to inject malicious code into the query. In our case, the worst that is likely is that someone will accidentally stick garbage in there and it'll fail to parse but not cause damage.

Rarely it is still necessary to use f-string formatting when dynamically setting table or column names rather than values, but that's uncommon. I think there's one example somewhere in this code.

src/cascade/input_data/db/bundle.py
alec-deason: Yep

src/cascade/input_data/db/bundle.py
alec-deason: I don't want to. We can if it really bothers people but it's totally unnecessary and ugly.

src/cascade/input_data/db/bundle.py
alec-deason: It's something you want to be conscious of but I think there are times where it's useful. For example, this case where I'm transforming a value several times and I want to make the steps clear (ie. not have a huge, complicated composed function) but I don't need or want to keep references to the intermediate versions of the data.

I don't know why Code Complete would make a hard rule about this, that seems dumb and you can tell whoever the mysterious person that gave you that feed back that I said so. Not doing it has several advantages: I have fewer names to keep track of. I don't have to think of new names for the intermediate values which would never be used anyway. The garbage collector may now be free to reap the old version of the value which it could not do if I'd changed names.

If I were reusing the same variable name for completely unrelated values then I agree, that seems like bad practice.

PR #5, opened 2018-07-26
Alec/feature/integrate drews dismod db wrapper

alec-deason:
The main difference between this and Drew's initial version (https://stash.ihme.washington.edu/projects/DIS/repos/cascade_at/pull-requests/69/overview) is that it stores data in memory until it is explicitly flushed and it updates the names of the tables to what I think is currently the correct names.


src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+This describes the tables in the sqlite file that Dismod reads.
+"""
+import enum
+import logging
+
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy import Column, Integer, String, Float, Enum, ForeignKey
+
+
+LOGGER = logging.getLogger("cascade_at.dm_file_format.dmtables")
```
nrhawkins: Maybe we want to change the "cascade_at..." part?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+
+    mulcov_id = Column(Integer(), primary_key=True)
+    mulcov_type = Column(Enum(MulCovEnum), nullable=False)
+    rate_id = Column(None, ForeignKey("rate.rate_id"), nullable=True)
+    """Determines the rate that this covariate and multiplier affects.
+    If mulcov_type is of type meas_value or meas_std, this must be null."""
+    integrand_id = Column(None, ForeignKey("integrand.integrand_id"), nullable=True)
+    covariate_id = Column(None, ForeignKey("covariate.covariate_id"), nullable=False)
+    smooth_id = Column(None, ForeignKey("smooth.smooth_id"), nullable=True)
+    """If this is null, the covariabe multiplier is always zero and no
```
nrhawkins: covariabe -> covariate

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+
+    option_id = Column(Integer(), primary_key=True)
+    option_name = Column(String(), unique=True)
+    option_value = Column(String())
+
+
+class Constraint(Base):
+    """Greg shows this table, but it's not in Brad's docs."""
+
+    __tablename__ = "constraint_table"
```
nrhawkins: Should this just be called "constraint" or would that be a problem?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+    time_lower = Column(Float())
+    time_upper = Column(Float())
+
+
+class CascadeOption(Base):
+    """
+    Greg shows this table, but it's not in Brad's docs.
+    """
+
+    __tablename__ = "cascade_option_table"
```
nrhawkins: Should this just be called "cascade_option"?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+    """
+
+    __tablename__ = "fit_var_table"
+    __readonly__ = True
+
+    fit_var_id = Column(Integer(), primary_key=True)
+    fit_var_value = Column(Float())
+    residual_value = Column(Float())
+    residual_dage = Column(Float())
+    redisual_dtime = Column(Float())
```
nrhawkins: redisual_dtime -> residual_dtime

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+    data_id = Column(None, ForeignKey("data.data_id"))
+
+
+class DependVar(Base):
+    """
+    Output, created by the depend command. Diagnostic to see whether the model
+    has more variables than necessary.
+    """
+
+    __tablename__ = "depend_var_table"
```
nrhawkins: Maybe just "depend_var"?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+
+
+class FitVar(Base):
+    """
+    Output, contains maximum likelihood estimate for the model variables
+    corresponding to the data table measured value. A new fit var table
+    is created each time the fit command runs.
+    """
+
+    __tablename__ = "fit_var_table"
```
nrhawkins: Maybe just "fit_var"?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+
+
+class FitDataSubset(Base):
+    """
+    Output, compares the model and data for the model variables corresponding
+    to a fit command. A new fit data subset table is created each time
+    the fit command runs.
+    """
+
+    __tablename__ = "fit_data_subset_table"
```
nrhawkins: Maybe just "fit_data_subset"?

src/cascade/dismod/db/metadata.py
```diff
@@ -0,0 +1,519 @@
...
+
+class SampleIndex(Base):
+    """
+    Output, The sample command creates this table with one optimal estimate
+    of the model variable. For each valid simulate index in the simulate
+    table, there is an equal sample index in the table with the optimal
+    variables corresponding to the measurement.
+    """
+
+    __tablename__ = "sample_index_table"
```
nrhawkins: Maybe just "sample_index"?

src/cascade/dismod/db/metadata.py
alec-deason: I have fallen into my own trap. This is why I like `logging.getLogger(__name__)`

src/cascade/dismod/db/metadata.py
alec-deason: I'm going to blame that one on Drew. Usually typos are my fault so I have to lean into it when they aren't.

src/cascade/dismod/db/metadata.py
alec-deason: Gah. Yes. I fixed one problem with these names which seems to have come out of issues with the documentation, but this is another that I didn't notice. We don't control these table names, they're Brad's. However this one (and the ones you point out below) are wrong. The `_table` suffix shouldn't be one there.

nrhawkins: ok, I stopped doing it after awhile, I am sure you noticed.

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
+"""
+Creates a file for Dismod-AT to read.
```
nrhawkins: I liked reading: "stores data in memory until it is explicitly flushed" , would it be useful to put that in the docstring, or is that too "how"?

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    Responsible for creation of a Dismod-AT file.
+
+    This class checks the type of all columns. It doesn't check that the
+    model is correct::
+
+        engine = _get_engine(None)
+        dm = DismodFile(engine, {"col": float}, {})
+        dm_file.add("time", pd.DataFrame({"time": [1997, 2005, 2017]}))
+        time_df = dm_file.time
```
nrhawkins: Is this correct example usage too?  something like,  dm_file.age = age_df

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+
+        is_new = table_name not in self._table_hash
+        if not is_new:
+            is_changed = not self._table_hash[table_name].equals(table_hash)
+        else:
+            is_changed = False
+
+        return is_new or is_changed
+
+    def flush(self):
```
nrhawkins: I wouldn't mind if one of the alternative names is used.   But, also can live with it.

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        """Writes any data in memory to the underlying database. Data which has not been changed since
+        it was last written is not re-written.
+        """
+        with self.engine.connect() as conn:
```
nrhawkins: down below, I see:    with self.engine.connect() as connection:

just wondering, if we like one more than the other?

src/cascade/dismod/db/metadata.py
```diff
@@ -98,21 +99,21 @@ class Node(Base):
 
     node_id = Column(Integer(), primary_key=True)
     node_name = Column(String(), nullable=False, unique=True)
-    parent = Column(Integer, nullable=True)  # Parent is an id in _this_ table.
+    parent = Column(None, ForeignKey("node.node_id"), nullable=True)  # Parent is an id in _this_ table.
```
adolgert: Warning: These can complicate the order of loading the tables. If you have foreign keys, you need to look at the dependency chain and insert tables in order.

src/cascade/dismod/db/wrapper.py
alec-deason: Hmm. We should be consistent. I'm going to change `conn` to `connection` because I strongly prefer complete words over cryptic abbreviations.

src/cascade/dismod/db/__init__.py
```diff
@@ -0,0 +1,2 @@
+class DismodFileError(Exception):
```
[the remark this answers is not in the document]
alec-deason: What do you mean by that? Are you saying this should inherit from a more general cascade exception? I agree with that. Or are you saying there's something that needs to be changed with the hierarchy in this section of code?

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+"""
+import logging
+
+import pandas as pd
+from sqlalchemy import create_engine
+from sqlalchemy.sql import select
+from sqlalchemy.exc import OperationalError
+
+from .metadata import Base, add_columns_to_avgint_table, add_columns_to_data_table, DensityEnum
+from . import DismodFileError
```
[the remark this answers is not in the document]
alec-deason: Yeah, I think if you're having trouble with this notation it's an indication that you're doing something wrong. But if we find a case where that's not true then we can change it for sure.

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def _is_dirty(self, table_name):
+        """Tests to see if the table's data has changed in memory since it was last loaded from the database.
+        """
+
+        if table_name not in self._table_data:
+            return False
+
+        table = self._table_data[table_name]
+        table_hash = pd.util.hash_pandas_object(table)
```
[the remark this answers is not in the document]
alec-deason: Meh, I'm not super happy with this. But the fact that the DataFrames are mutable it's tricky to track whether they are dirty without writing a complex proxy on top of them. This is the CPU and memory intensive way of avoiding that code complexity. But I doubt it will cause any trouble in practice.

tests/dismod/db/test_wrapper.py
```diff
@@ -0,0 +1,81 @@
...
+import pytest
+
+from sqlalchemy.exc import StatementError
+
+from cascade.dismod.db.wrapper import DismodFile, _get_engine
+
+
+@pytest.fixture
+def engine():
+    return _get_engine(None)
```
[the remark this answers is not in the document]
alec-deason: I have zero problems calling internal APIs in tests. Is that something you're worried about?

tests/dismod/db/test_wrapper.py
```diff
@@ -0,0 +1,81 @@
...
+    dm_file.integrand = pd.DataFrame({"integrand_name": ["prevalence"]})
+
+    return dm_file
+
+
+def test_wrong_type(base_file):
+    ages = pd.DataFrame({"age": np.array(["strings", "for", "ages"])})
+    base_file.age = ages
+    with pytest.raises(StatementError):
+        base_file.flush()
```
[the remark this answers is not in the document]
alec-deason: We could do explicit validation elsewhere, but it's tricky. Pandas dataframes don't (as far as I know) provide a way to hook behavior onto column mutations. That means that I can't really know what the contents of a given DataFrame are until I actually check which I only do at flush time. Like the dirty bit stuff above, you could solve this by writing a complex proxy. Or changing the API such that you had to explicitly set a whole DataFrame at a time so there was a place to do these checks. Something like this:

data_frame = dbfile.get_data("key")
dbfile.set_data("key", data_frame)

That makes this validation stuff much easier, but it's a less pleasant API. I think I like the current version, even though it throws late rather than early in this case.

src/cascade/dismod/db/wrapper.py
```diff
@@ -0,0 +1,153 @@
...
+        it was last written is not re-written.
+        """
+        with self.engine.connect() as conn:
+            for table_name, table in self._table_data.items():
+                if self._is_dirty(table_name):
+                    if hasattr(table, "__readonly__") and table.__readonly__:
+                        raise DismodFileError(f"Table '{table_name}' is not writable")
+
+                    table_definition = self._table_definitions[table_name]
+                    table.to_sql(
```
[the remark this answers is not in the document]
alec-deason: Can I actually make an error that's clearer than the one sqlalchemy will throw? I guess I could hide the fact that it's a SQL database under the hood and talk only in python terms. That's compelling.

tests/dismod/db/test_wrapper.py
```diff
@@ -0,0 +1,81 @@
...
+def test_is_dirty__not_yet_read(base_file):
+    assert not base_file._is_dirty("foo_bar")
+
+
+def test_dmfile_read(base_file, engine):
+    ages = base_file.age
+    times = base_file.time
+    base_file.flush()
+
+    dm_file2 = DismodFile(engine, {"howdy": float}, {"there": int})
```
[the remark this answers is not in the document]
alec-deason: Between sqlalchemy and sqlite, they seem to do the right things. It could be that by not making that explicit we're exposed to problems in the future. Dunno.

src/cascade/dismod/db/metadata.py
alec-deason: Yep. But without this you have no guarantees that your keys match. Especially with the possibility of people mucking with tables in pandas outside the DB that seems like something we want to be careful with. But you're right about table order, I haven't made that explicit.

tests/dismod/db/test_wrapper.py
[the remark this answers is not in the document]
adolgert: Yup. That's what I was worried about. It seems like a sign that maybe engine creation doesn't belong shoved at the top of another file. I can't tell yet where it goes, though, so might as well let it happen for now.

tests/dismod/db/test_wrapper.py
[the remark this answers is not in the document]
alec-deason: I'll admit, I didn't think about this at all while working on the code.

tests/dismod/db/test_wrapper.py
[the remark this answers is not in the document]
adolgert: Let's not mess with the Pandas api but focus on throwing a beautiful error on flush.

src/cascade/dismod/db/wrapper.py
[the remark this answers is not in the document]
alec-deason: While I'm in here, I'm also going to make the whole flush transactional.

tests/dismod/db/test_wrapper.py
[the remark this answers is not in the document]
alec-deason: Yeah, I think the way we're handling the engine is slightly weird but I don't have a better suggestion now. In general though I don't think reaching under the public API when setting up tests is an indication of any real problem. Tests are inherently internal.

src/cascade/dismod/db/wrapper.py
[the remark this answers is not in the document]
alec-deason: Wobble. Maybe this isn't as bad as I think...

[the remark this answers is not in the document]
adolgert: Wobble, hah. Larger task? Here, I defer to your judgment on the size of the task and whether it should be in this PR. We could happily put this into the backlog.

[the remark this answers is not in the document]
alec-deason: I don't know if this was worth 2+ hours but there's a validator now that raises meaningful errors.

src/cascade/dismod/db/wrapper.py
```diff
@@ -27,6 +30,50 @@ def _get_engine(file_path):
...
+                    actual_type = type(data[column_name][0])
+                    correct = np.issubdtype(actual_type, np.str_)
+
+                if not correct:
+                    raise DismodFileError(
+                        f"column '{column_name}' in data for table '{table_definition.name}' must be string"
+                    )
+        elif not (column_definition.primary_key or column_definition.nullable):
+            raise DismodFileError(f"Missing column in data for table '{table_definition.name}': '{column_name}'")
+        columns_checked.add(column_name)
```
adolgert: Pretty.

tests/dismod/db/test_wrapper.py
```diff
@@ -79,3 +83,73 @@ def test_reading_modified_columns(base_file, engine):
...
+
+
+class DummyTable(DummyBase):
+    __tablename__ = "test_table"
+    primary_key_column = Column(Integer(), primary_key=True)
+    integer_column = Column(Integer())
+    float_column = Column(Float())
+    nonnullable_column = Column(Integer(), nullable=False)
+    string_column = Column(String())
+    enum_column = Column(Enum(enum.Enum("Bee", "bumble honey carpenter wool_carder")))
```
adolgert: Good idea to make a dummy table.

PR #7, opened 2018-07-30
Making realtime reading from subprocess pipes work as expected

alec-deason:


src/cascade/executor/dismod_runner.py
```diff
@@ -134,6 +127,22 @@ def reduce_process_priority():
...
+
+    Args:
+        pipe: The pipe to read from
+        result: a list to accumulate the output into
+        callback: a callable which will be invoked each time data is read from the pipe
+    """
+    while not pipe.at_eof():
+        text = yield from pipe.read(2 ** 16)
+        text = text.decode("utf-8")
+        result.append(text)
```
adolgert: This is a choice to keep the whole stdout in memory and to send everything-so-far to the callback.
The goal of this callback, currently, is to append to a text file that can be read, and refreshed, through EpiViz and a browser. It might work better to send up to the last line, but that wouldn't be decided here, I suppose.

src/cascade/executor/dismod_runner.py
```diff
@@ -153,51 +162,46 @@ def run_and_watch(command, single_use_machine, poll_time):
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            universal_newlines=True,
-            preexec_fn=pre_execution_function
-         )
-    except ValueError as ve:
-        raise Exception(
-            f"Dismod called with invalid arguments {ve}"
+        sub_process = yield from asyncio.subprocess.create_subprocess_exec(
+            *command, stdout=asyncio.subprocess.PIPE, stderr=asyncio.subprocess.PIPE, preexec_fn=pre_execution_function
```
adolgert: Took out universal newlines? It was there... why? There was a reason. I should have left a comment about that, and now I can't remember. It turns all line endings into \n, whether they come from PC or from Mac or Linux. Maybe I just liked that idea.

src/cascade/executor/dismod_runner.py
```diff
@@ -153,51 +162,46 @@ def run_and_watch(command, single_use_machine, poll_time):
...
+    out_list = []
+    err_list = []
 
-    outs, errs = sub_process.communicate()
-    MATHLOG.debug(outs)
-    out_list.append(outs)
-    err_list.append(errs)
+    loop = asyncio.get_event_loop()
+    std_out_task = loop.create_task(_read_pipe(sub_process.stdout, out_list, lambda text: MATHLOG.debug(text)))
+    std_err_task = loop.create_task(_read_pipe(sub_process.stderr, err_list, lambda text: MATHLOG.error(text)))
```
adolgert: Separate for stderr and stdout. Makes sense. This code is pretty compact. If it passes tests, then I'm good with this level of complexity.

src/cascade/executor/dismod_runner.py
alec-deason: I'm preserving the existing behavior. Rather than having the mutable accumulator and the callback, we could just have the callback which sends to the logs and/or does something else. I don't fully understand what's happening with this code so I wanted to keep my changes local.

src/cascade/executor/dismod_runner.py
alec-deason: From the asyncio docs: "The universal_newlines parameter is not supported (only bytes strings are supported)"

That said, I'm not particularly interested in supporting Windows unless you think it's worth putting effort into.

alec-deason:
The way to think about pipes is as if they were files (which they basically are). One process writes to them (`ls` in the test code) and the other process reads (the parent python program). They can be bidirectional too, but we aren't using them that way here. The problem arises because the natural (ie. the most basic tools the operating system provides) way of reading from files is blocking which means that if the file isn't marked with an EOF (end of file) then your process will sit and wait until someone writes data to the file so you can read it. Pipes get EOF once the process writing to them exits. That means that normally you can't get data out of the pipe until the end of the process you've launched. But for long lived processes that's bad because you'd really like to have access to intermediate outputs in real time.

There are a bunch of ways of solving the problem. The OS provides several tools (`select`, `poll` and some others) for checking if data is available so you know when to do your reads. You can build a real time pipe reader using those but I don't understand them very well so I chose not to. They are the traditional way of doing it. Recently (10 years? Probably more) people have started abstracting those low level tools up into non-blocking IO systems which give you different ways of talking about these problems. `asyncio` is an example of a non-blocking IO system and it happens to be part of the python standard library, so awesome.

`asyncio` works by giving you the ability to schedule coroutines which are basically threads but managed by the python VM rather than the host OS. Coroutines never run concurrently, meaning there's only ever one of them executing at a time, but they can release control back to the VM to allow other coroutines to run (this happens whenever you see a `yield` in that code). This is a model called cooperative scheduling as opposed to preemptive scheduling where the host environment forces threads to give up control so other code can run, that's what the OS does with regular threads and processes.

So in this code I create three coroutines. One of them runs the subprocess until it finishes and the other two attach to the pipes and read data out until the pipes close. Once I've launched all three, I go back and wait for each of them to finish. Under the hood the `asyncio` system is managing the reads from the pipes for me so I don't have to muck around with `select` or whatnot. In my code I get to treat reads like regular blocking reads because I don't care if one of the coroutines blocks, the others can keep running if they need to.

Anyway, I don't know if that makes what's going on any clearer. There's a bunch of concepts there which is why Drew is talking about "code of this complexity" but it's all worth understanding. Though we probably won't need to do much with this kind of IO elsewhere in this project.

PR #8, opened 2018-07-30
added sphinx libraries to setup.py extra libs

adolgert:
This was just missing from the library dependencies.


alec-deason:
Looks good. I'd kind of like to see us testing that the docs actually build under the CI. I think it's possible to do this with a little fiddling under tox/travis. Specifically Travis's jobs.include configuration which, I think, is designed for multiple groups of tests like this (unit tests, doc tests, etc all as seperate jobs) but it's poorly documented. I'm happy to muck with it and figure out how it works.

PR #9, opened 2018-07-30
Feature/dismodat 48 load csmr to t3

nrhawkins:


src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+    return exists == 1
+
+
+def _get_csmr_data(execution_context):
+
+    model_version_id = execution_context.parameters.model_version_id
+    add_csmr_cause = execution_context.parameters.add_csmr_cause
+    parent_loc = execution_context.parameters.drill
+
+    csmr_measure_id_deaths = 1  # measure type = deaths (1)
```
alec-deason: Let's move all three of these to constants in the `__init__.py` of this package. There's the beginning of a mapping from measure names to measure ids in bundle.py which should probably also move into the top level. Eventually we'll want something better because having these kinds of magic constants is super dangerous but I'm doing it in my code too because I haven't thought of a better pattern yet.

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+    csmr_metric_id_rate = 3  # metric type = per capita rate
+    csmr_gbd_round_id = 5
+
+    csmr_keep_cols = ["year_id", "location_id", "sex_id", "age_group_id",
```
alec-deason: No need for the `csmr_` prefix, everything here is related to `csmr`

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+        sex_id="all",
+        gbd_round_id=csmr_gbd_round_id
+        )[csmr_keep_cols]
+
+    csmr["model_version_id"] = model_version_id
+
+    csmr_columns = ["model_version_id"] + csmr_keep_cols
+    csmr = csmr[csmr_columns]
+
+    csmr.rename(columns={"val": "mean"}, inplace=True)
```
alec-deason: Let's avoid inplace operations for the sake of consistency. In general they're no cheaper.

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+            model_version_id,
+            year_id,
+            location_id,
+            sex_id,
+            age_group_id,
+            mean,
+            lower,
+            upper
+        ) VALUES (
+            {model_version_id}, {", ".join(["%s"]*32)}
```
alec-deason: There are not 32 parameters here. Make sure you try this out on the actual database to flush problems like this. I know I keep harping on how we can't have database access in the unit tests but we do need to actually test stuff against the database. Soon we'll have integration and/or smoke tests that will automate that but for now make sure you're doing it by hand.

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+    """
+
+    cursor.executemany(insert_query, csmr_data.values())
+
+    CODELOG.debug(f"uploaded {len(csmr_data)} lines of csmr data")
+
+
+def load_csmr_to_t3(execution_context) -> bool:
+    """
+    Upload to t3_model_version_csmr if the user requested that and if
```
alec-deason: What does "if the user requested that" refer to?

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+    it's not already there.
+    """
+
+    model_version_id = execution_context.parameters.model_version_id
+    add_csmr_cause = execution_context.parameters.add_csmr_cause
```
alec-deason: Why would we be calling this function if we didn't want to upload? If there's some conditional it'll be in the calling code.

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+        c.execute(query, args={"model_version_id": model_version_id})
+        exists = c.fetchone()[0]
+
+    return exists == 1
+
+
+def _get_csmr_data(execution_context):
+
+    model_version_id = execution_context.parameters.model_version_id
+    add_csmr_cause = execution_context.parameters.add_csmr_cause
```
alec-deason: I'm not sure what `add_csmr_cause` means. I believe there will only ever be one cause per model so we can probably just call this `cause_id`

src/cascade/input_data/db/csmr.py
```diff
@@ -0,0 +1,136 @@
...
+    parent_loc = execution_context.parameters.drill
```
alec-deason: Why `drill`? I would call this `location_id` or `node_id`. The context will (I think) only ever be in reference to one node at a time.

src/cascade/input_data/db/csmr.py
nrhawkins: This is a user setting in the EpiViz interface, which goes to the JSON string.
add_csmr_cause

src/cascade/input_data/db/csmr.py
nrhawkins: Ok.  The ticket mentioned the conditional.

src/cascade/input_data/db/csmr.py
nrhawkins: ok.  I was thinking of the JSON settings string.

src/cascade/input_data/db/csmr.py
alec-deason: Ok. I think that's a mistake in the ticket. We may need the conditional but not at this level.

src/cascade/input_data/db/csmr.py
nrhawkins: How do you mean this?  Something other than this, I guess.

(20180722) [nhawkins@cn497 db]$ more __init__.py 
MODEL_VERSION_ID = execution_context.parameters.model_version_id
CAUSE_ID = execution_context.parameters.add_csmr_cause
PARENT_LOC = execution_context.parameters.location_id

src/cascade/input_data/db/csmr.py
nrhawkins: I was looking at bundle.py's:  
def _upload_bundle_data_to_tier_3(cursor, model_version_id, bundle_data):

in which I see this:
cursor.executemany(insert_query, bundle_data.values())

Is bundle_data df?  I thought it was.  I must be missing something.

In a python terminal window, I created a df called x and tried to call values().
>>> type(x)
<class 'pandas.core.frame.DataFrame'>
>>> x.values()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'numpy.ndarray' object is not callable

alec-deason: Oh, dear. I have failed to do exactly the thing I was telling you to do. The problem is that the code I wrote doesn't work. It should be `.values` not `.values()`

src/cascade/input_data/db/csmr.py
alec-deason: I'm talking about csmr_measure_id_deaths, csmr_metric_id_rate and csmr_gbd_round_id. Those are constants which are defined by the external database but which we need to know here and likely elsewhere as well so they should be in a more central place. I suggest `cascade.input_data.db`

In the long term we should have mappings for all measure_ids (that we care about) and all metric_ids (that we care about). We'll never be working on more than one round at a time (I hope).

Did you look at the measure_id mapping in bundle.py? That's one possible way of approaching the more general version of this. I suggest you take that, move it to `cascade.input_data.db` and add the death measure to it then create a parallel thing for the metric_ids and just a regular constant for the round_id.

src/cascade/input_data/db/csmr.py
adolgert: Yes, the drill, for the example we've seen, did contain the parent. I'm guessing there will be a translation step from the settings to a more generic parent-child tree. The execution context might contain both parent and children, or it could have a view of the tree, if that's what we need. Unclear now.

tests/conftest.py
```diff
@@ -16,4 +17,4 @@ def mock_ezfuncs(mocker):
 
 @pytest.fixture
 def mock_database_access(mock_ezfuncs):
-    return {"cursor": mock_ezfuncs.get_connection().cursor(), "connection": mock_ezfuncs.get_connetion()}
```
alec-deason: Bleh. Thanks for catching that.

PR #10, opened 2018-07-31
Feature/dismodat 239 download mortality data

alec-deason:
This is just copying over the queries from the old importer. It's difficult to test until the PRs for freezing data merge so it may need some work at that point.


src/cascade/input_data/db/bundle.py
```diff
@@ -196,7 +196,7 @@ def _upload_bundle_data_to_tier_3(cursor, model_version_id, bundle_data):
     )
     """
 
-    cursor.executemany(insert_query, bundle_data.values())
+    cursor.executemany(insert_query, bundle_data.values)
```
nrhawkins: Looks like df.to_sql needs: 
    con : sqlalchemy.engine.Engine or sqlite3.Connection: 
rather than:
    db_tools.ezfuncs.get_connection(conn_def, raw=False, connectable=True)
        Returns: | sqlalchemy.pool._ConnectionFairy or db_tools.mysqlapis.MySQLConnection:
I made a quick attempt at calling ezfuncs.get_engine() instead of ezfuncs.get_connection(), and I got an error that user dbview is not allowed to insert into table t3_model_version_csmr


with a cursor, I unsuccessfully tried each of these individually as inputs to executemany:
    csmr_data_values = list(csmr_data.itertuples(index=False))
    csmr_data_values = [tuple(x) for x in csmr_data.values] 
    For this last one, it doesn't like the numpy types, float or int., 
    eg., AttributeError: 'numpy.int64' object has no attribute 'translate'

src/cascade/input_data/db/mortality.py
```diff
@@ -0,0 +1,76 @@
...
+            SELECT
+                location_id,
+                year_id as time_lower,
+                age_group_id,
+                sex_id as x_sex,
+                mean as meas_value,
+                lower as meas_lower,
+                upper as meas_upper
+            FROM
+                epi.t3_model_version_csmr t3_csmr
```
nrhawkins: I am wondering why there are 2 items on the FROM line?
epi.t3_model_version_csmr 
t3_csmr

src/cascade/input_data/db/mortality.py
```diff
@@ -0,0 +1,76 @@
...
+                upper as meas_upper,
+                standard_error
+            FROM
+                epi.t3_model_version_emr t3_emr
+            WHERE model_version_id = %(model_version_id)s
+            AND t3_emr.outlier_type_id = 0
+    """
+
+    with connection(execution_context) as c:
+        data = pd.read_sql(query, c, params={"model_version_id": model_version_id})
```
nrhawkins: I am wondering which of these c is: (from pandas.read_sql doc), a _ConnectionFairy?
    con : SQLAlchemy connectable (engine/connection) or database string URI or DBAPI2 connection (fallback mode)
        Using SQLAlchemy makes it possible to use any DB supported by that library. 
        If a DBAPI2 object, only sqlite3 is supported.

alec-deason: You can answer this question either by tracing the code back to where the connection get's created (there's a description of what it is in the ezfuncs code)  or by using a debugger to examine the live object. If you aren't familiar with debuggers I can show you how to work with the python one. It's a useful tool to have access to.

src/cascade/input_data/db/mortality.py
alec-deason: They're creating an alias for the table name. If you look at the rest of the query you can see them refer to it by the shorter version.

nrhawkins: I don't see it being used in lines 36-48 which is to the end of the query, except in line 46.

src/cascade/input_data/db/mortality.py
nrhawkins: I may have another question, that I am not asking here.   I did look at the type previously, and saw ConnectionFairy.

src/cascade/input_data/db/bundle.py
nrhawkins: PR #14 has the bundle code fiddles for this one, as you know, just saying

src/cascade/input_data/db/mortality.py
nrhawkins: Just saying, I noticed in the file a couple places where the FROM line has the two names, but the shorter name doesn't appear anywhere else in the query.  I did see in at least one query the shorter name is used.  Probably not a big deal.

alec-deason: Ah, you're right. It's superfluous here. I must have been looking at one of the other queries.

PR #11, opened 2018-07-31
Feature/dismodat 50 load mtall to t3

skspoon:
WIP:  The asdr.py passed the mock tests in the test_asdr.py,  Unit testing with data will be needed.


docs/input_data.rst
```diff
@@ -0,0 +1,7 @@
+.. _input_data:
+
+Input Data Functionality
+========================
+
+.. automodule:: cascade.input_data.db.asdr
+    :members:
```
nrhawkins: I guess we need to add similar lines for bundle and csmr.  

In csmr, I had a line which stopped at db (though it also erroneously had "core" in front of it, rather than "input_data").

PR #12, opened 2018-08-01
Feature/build the docs

adolgert:
Small changes that made the docs build on buildthedocs.
Note that we now have a separate requirements.txt in the docs directory which will become out of sync with actual requirements. It has to be just enough to load the files.

I only spent more time on this b/c James helped. The trick was making the repo not zip safe and telling conf.py that our source was in ../src. Docs are now available here:
https://cascade.readthedocs.io/en/latest/


readthedocs.yml
```diff
@@ -0,0 +1,5 @@
+build:
+  image: latest
+python:
+  version: 3.6
+  setup_py_install: true
```
alec-deason: Does pip_install let us get away without the requirements.txt?

setup.py
```diff
@@ -18,4 +18,12 @@
...
         ]
     },
+    zip_safe=False,
+    classifiers=[
+        "Intendend Audience :: Science/Research",
+        "Development Status :: 3 - Alpha",
+        "License :: OSI Approved :: MIT License",
+        "Programming Language :: Python :: 3.6",
+        "Topic :: Scientific/Engineering :: Statistics",
+    ],
```
alec-deason: Yesyes

readthedocs.yml
alec-deason: There's also `extra_requirements` if one of the optional requirements was the problem.

docs/requirements.txt
```diff
@@ -0,0 +1,36 @@
+alabaster==0.7.11
```
alec-deason: :(:(:( But I know you tried to find a way not to need this...

adolgert: haha. they import it. Maybe it's unused? sphinx wants it. It wasn't me!!!

alec-deason: Oh, I was talking about the whole file. I just don't see a way to comment on the file itself.

readthedocs.yml
alec-deason: Ok. Whatever. Reading through their docs and forums it really seems like we shouldn't need anything other than the setup.py but it's certainly not worth spending more time on right now.

PR #13, opened 2018-08-02
writes dismod file in order dictated by foreign key dependencies

adolgert:
Even sqlite3 won't let you write tables with foreign keys that are written out-of-order because it needs to check the keys. This PR adds a function to the file wrapper that finds all foreign keys and sorts the tables by their dependency on each other.

- it adds a library dependency to the networkx library, because that has a trustworthy sorting algorithm. If you want to rewrite that later, it's a nice exercise and limits our dependencies.
- it takes the foreign key off of the node_id.parent because that's a self-key. That will _never_ work with pandas updates to the database.
- adds docs for the Dismod file to the references section, which we should augment over time.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ setup(
     packages=PEP420PackageFinder.find("src"),
     package_data={"cascade.executor": ["data/*.toml"]},
     package_dir={"": "src"},
-    install_requires=["numpy", "pandas", "scipy", "toml", "sqlalchemy"],
+    install_requires=["numpy", "pandas", "scipy", "toml", "sqlalchemy", "networkx"],
     extras_require={
         "testing": ["hypothesis", "pytest", "pytest-mock"],
         "documentation": ["sphinx", "sphinx_rtd_theme", "sphinx-autobuild", "sphinxcontrib-napoleon"],
--- a/src/cascade/dismod/db/wrapper.py
+++ b/src/cascade/dismod/db/wrapper.py
@@ -7,6 +7,8 @@ to create it and add tables.
 import logging
 from enum import Enum
 
+from networkx import DiGraph
+from networkx.algorithms.dag import lexicographical_topological_sort
 import pandas as pd
 import numpy as np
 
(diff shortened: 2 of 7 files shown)


setup.py
```diff
@@ -6,7 +6,7 @@
     packages=PEP420PackageFinder.find("src"),
     package_data={"cascade.executor": ["data/*.toml"]},
     package_dir={"": "src"},
-    install_requires=["numpy", "pandas", "scipy", "toml", "sqlalchemy"],
+    install_requires=["numpy", "pandas", "scipy", "toml", "sqlalchemy", "networkx"],
```
alec-deason: This is fine. I suspect we'll end up wanting it for other things anyway. We're operating on a graph-like object.

src/cascade/dismod/db/wrapper.py
```diff
@@ -74,6 +76,35 @@ def _validate_data(table_definition, data):
...
+    Args:
+        schema (sqlalchemy.metadata): Metadata about the tables.
+        tables_to_write (iterable[str]): An iterable of table names as string.
+
+    Returns:
+        Iterates through the tables in an order that is safe for writing.
+    """
+    dependency_graph = DiGraph()
+
+    if len(set(tables_to_write) - set(schema.tables.keys())) > 0:
```
alec-deason: You don't need the `len` here. Non-empty sets are truthy.

src/cascade/dismod/db/wrapper.py
```diff
@@ -74,6 +76,35 @@ def _validate_data(table_definition, data):
...
+        raise ValueError("Asking to write tables not in schema")
+
+    for scan_name, scan_table in schema.tables.items():
+        dependency_graph.add_node(scan_name)
+        for foreign_key in scan_table.foreign_keys:
+            target_name = foreign_key.target_fullname.split(".")[0]
+            dependency_graph.add_edge(target_name, scan_name)
+
+    # Use the full lexicographical sort because it makes testing deterministic.
```
alec-deason: As opposed to what?

adolgert: There is a purely topological sort, but then the order of tables won't be deterministic. The lexicographical (within toplogical constraints) is deterministic.

setup.py
adolgert: fixed

src/cascade/dismod/db/wrapper.py
alec-deason: I see. That makes sense.

PR #14, opened 2018-08-02
Fiddle the bundle code

alec-deason:
My initial work on bundle uploading didn't actually work. This one works better. We need a way to write effective integration tests to better catch these problems (https://jira.ihme.washington.edu/browse/DISMODAT-268).

### DIFF
--- a/src/cascade/input_data/db/bundle.py
+++ b/src/cascade/input_data/db/bundle.py
@@ -285,7 +285,7 @@ def _normalize_bundle_data(data):
 def _covariate_ids_to_names(execution_context, study_covariates):
     """Convert study_covariate_ids to canonical study covariate names
     """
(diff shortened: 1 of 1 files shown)


src/cascade/input_data/db/bundle.py
```diff
@@ -196,7 +196,8 @@ def _upload_bundle_data_to_tier_3(cursor, model_version_id, bundle_data):
     )
     """
 
-    cursor.executemany(insert_query, bundle_data.values())
+    bundle_data = bundle_data.where(pd.notnull(bundle_data), None)
+    cursor.executemany(insert_query, bundle_data.values.tolist())
```
nrhawkins: I used and tested both of these for the csmr task.

PR #15, opened 2018-08-07
Feature/dismodat 265 smooth grids

alec-deason:
This is a set of tools for creating grids of priors. It is oriented towards the case when a modeler or application is setting up the sparse priors at the beginning of a model. The case where someone is looking at the dense prior grids post-fitting will be handled by the point-query method here along with some code to read from the underlying database which is coming in the next PR along with code to write this stuff out to the database.


alec-deason:
Aside from correctness, the thing I'm most interested in here is clear terminology. I want this to make sense both to application developers and to research oriented end users (and to Brad). So, please nit-pick at names and things as well as the actually structure.

docs/refmanual.rst
```diff
@@ -6,4 +6,5 @@ Cascade-AT Reference Manual
    :maxdepth: 2
 
    core
+   model
```
nrhawkins: Should these be listed in alphabetical order, or does the order start at core and go out from there, somehow?

alec-deason: Hmm. I didn't even think about order. I was just cramming things in willy nilly. I think you could make a case for either most general -> most specific or alphabetical as the correct order. Do we have an established pattern?

src/cascade/model/compartmental.py
```diff
@@ -9,8 +9,8 @@
 
 1.  Specify a disease process by making simple Python functions that
     return disease rates as a function of time. You can specify a set
-    of :math:`(\iota, \rho, \chi, \mu)`, or you can specify a set
-    of :math:`(\iota, \rho, \chi, \omega)`. We'll call the former the
+    of :math:`(\iota, \\rho, \chi, \mu)`, or you can specify a set
+    of :math:`(\iota, \\rho, \chi, \omega)`. We'll call the former the
```
nrhawkins: What is the double backslash adding?

src/cascade/model/compartmental.py
```diff
@@ -93,7 +93,7 @@ def lx(t):
 
 .. math::
 
-    P = \frac{C}{S+C}
+    P = \\frac{C}{S+C}
```
nrhawkins: What is the double backslash adding?

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+        raise ValueError("Upper age not in underlying grid")
+
+    if not _any_close(lower_time, grid.times, grid.snap_distance) and not isinf(lower_time):
+        raise ValueError("Lower time not in underlying grid")
+
+    if not _any_close(upper_time, grid.times, grid.snap_distance) and not isinf(upper_time):
+        raise ValueError("Upper time not in underlying grid")
+
+
+class _RegionView:
```
nrhawkins: When first reading this name, _RegionView, made me wonder - what is this?
Maybe add a docstring for this class.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+        else:
+            upper_time = self._time_slice.stop
+
+        return (lower_age, upper_age, lower_time, upper_time)
+
+    @property
+    def prior(self):
+        lower_age, upper_age, lower_time, upper_time = self._rectangle()
+        if lower_age != upper_age or lower_time != upper_time:
+            raise NotImplementedError("Currently only point queries are supported")
```
nrhawkins: Looks like a point means a single age OR a single time, just checking, is that the intention; i.e. doesn't have to be both a single age and a single time?

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+
+    if not _any_close(lower_time, grid.times, grid.snap_distance) and not isinf(lower_time):
+        raise ValueError("Lower time not in underlying grid")
+
+    if not _any_close(upper_time, grid.times, grid.snap_distance) and not isinf(upper_time):
+        raise ValueError("Upper time not in underlying grid")
+
+
+class _RegionView:
+    def __init__(self, parent, age_slice, time_slice):
```
nrhawkins: I found myself wondering what is a parent.

src/cascade/model/compartmental.py
alec-deason: It adds success! Without it sphinx fails because `\r` is a special thing so you have to escape the slash in order to get `\rho`

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+class PriorGrid:
+    """Represents priors for rectangular regions of an underlying age-time grid.
+
+    Args:
+        grid: The underlying age-time grid
+        hyper_prior: The prior for the priors in this grid
+
+    Examples:
+        >>> grid = AgeTimeGrid.uniform(age_start=0,age_end=120,age_step=5,time_start=1990,time_end=2018,time_step=1)
+        >>> d_time = PriorGrid(grid)
```
nrhawkins: Does "d" stand for density?

src/cascade/model/grids.py
alec-deason: Sure

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+        querying, more recent priors take precedence.
+        """
+        _validate_region(self._grid, lower_age, upper_age, lower_time, upper_time)
+        self._priors.append(((lower_age, upper_age, lower_time, upper_time), prior))
+
+    def _prior_at_point(self, age, time):
+        """Find the prior for a particular point on the age-time grid.
+        """
+        final_prior = None
+        for ((lower_age, upper_age, lower_time, upper_time), prior) in self._priors:
```
nrhawkins: This is similar to reversing the _priors and returning at the first one found.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,174 @@
...
+
+GRID_SNAP_DISTANCE = 1 / 365
+
+
+class AgeTimeGrid:
+    """Structure for storing age and times over which the model will estimate.
+    Also constructors for making standard age-time grids.
+    """
+
+    @classmethod
```
nrhawkins: I was reading that one purpose of using a class method involves inheritance.  If a class inherits from AgeTimeGrid,  and uniform() is called, that class will be returned rather than AgeTimeGrid.  I will read some more, just wondering as a I read through.

src/cascade/model/grids.py
alec-deason: So it is. That's certainly a simpler implementation. I'm going to switch to that.

src/cascade/model/grids.py
alec-deason: That's correct. I don't expect this to be used that way. I'm actually not sure why I didn't make it a separate factory function, that's what I'd normally do. I'm imagining that there might be several of these alternate constructors which is why I thought to attach it to the class. Dunno. Many ways to skin cats etc.

tests/model/test_grids.py
```diff
@@ -0,0 +1,72 @@
...
+
+@pytest.fixture
+def age_time_grid():
+    return AgeTimeGrid.uniform(age_start=0, age_end=120, age_step=5, time_start=1990, time_end=2018, time_step=1)
+
+
+def test_PriorGrid__development_target():
+    # Make a grid using a standard structure. Other options might be
+    # a constructor which makes a grid that's denser at extreme ages.
+    grid = AgeTimeGrid.uniform(age_start=0, age_end=120, age_step=5, time_start=1990, time_end=2018, time_step=1)
```
nrhawkins: Would it make sense to use the age_time_grid fixture inside this test?

nrhawkins: Would a denser grid also be created as a fixture?

alec-deason: You could. For these particular tests the exact structure of the grid doesn't really matter.

src/cascade/model/grids.py
nrhawkins: sure, thanks.  I am still learning my approach to PR reading.  OR is complicated.  :-)

tests/model/test_grids.py
nrhawkins: Ok, I was just noticing that you aren't using the existing fixture for this test.

alec-deason: Oh, yeah. I left it there because this was my target while I was writing out the implementation and I wanted to keep it as a mini-example.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,234 @@
+"""Support for dismod concepts that are expressed over grids: age-time, smoothings and value priors.
+"""
+from math import isclose, isinf
+
+import numpy as np
+
+GRID_SNAP_DISTANCE = 1 / 365
```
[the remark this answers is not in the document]
alec-deason: Ok. I'm not totally happy with how this is setup anyway. It's used inconsistently across the code I've written. Need to think about how to do this.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,234 @@
...
+
+    @property
+    def times(self):
+        return self._times
+
+    @property
+    def snap_distance(self):
+        return self._snap_distance
+
+    def to_age(self, query):
```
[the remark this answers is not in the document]
alec-deason: I'm using infinities to represent the space beyond the edge of the grid explicitly. Brad, if I understand correctly, does this implicitly.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,234 @@
...
+    if not _any_close(lower_time, grid.times, grid.snap_distance) and not isinf(lower_time):
+        raise ValueError("Lower time not in underlying grid")
+
+    if not _any_close(upper_time, grid.times, grid.snap_distance) and not isinf(upper_time):
+        raise ValueError("Upper time not in underlying grid")
+
+
+class _RegionView:
+    """Represents a slice into a PriorGrid and exposes a way to query or set
+    the prior over that slice.
```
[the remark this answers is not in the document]
alec-deason: Based on the conversation I had with Greg, in which he had to explain this to me several times because I was making the same assumption as you are, that is not correct. points in the smoothing grid represent the prior from that point forward (in both age and time) until the next prior, if there is one. That's why I chose to represent priors as rectangles. Now it's possible that I'm still misunderstanding what actually happens in which case you are correct and this doesn't make any sense.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,234 @@
...
+    Examples:
+        >>> grid = AgeTimeGrid.uniform(age_start=0,age_end=120,age_step=5,time_start=1990,time_end=2018,time_step=1)
+        >>> d_time = PriorGrid(grid)
+        >>> #Set a prior for the whole grid:
+        >>> d_time[:, :].prior = GaussianPrior(0, 0.1)
+        >>> #Set a prior for a band of ages
+        >>> d_time[0:15, :].prior = GaussianPrior(1, 0.01)
+        >>> #Or a single year
+        >>> d_time[:, 1995].prior = GaussianPrior(0, 3)
+    """
```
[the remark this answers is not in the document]
alec-deason: Meh. I find them to be too much trouble. I was just trying to give an example. Though if we're going to have examples in the docstrings they probably should get tested so they don't drift.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,234 @@
...
+        grid: The underlying age-time grid
+        hyper_prior: The prior for the priors in this grid
+
+    Examples:
+        >>> grid = AgeTimeGrid.uniform(age_start=0,age_end=120,age_step=5,time_start=1990,time_end=2018,time_step=1)
+        >>> d_time = PriorGrid(grid)
+        >>> #Set a prior for the whole grid:
+        >>> d_time[:, :].prior = GaussianPrior(0, 0.1)
+        >>> #Set a prior for a band of ages
+        >>> d_time[0:15, :].prior = GaussianPrior(1, 0.01)
```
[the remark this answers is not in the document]
alec-deason: I am treating slices on both dimensions as floating point years. I guess I don't have an example of using floats here, probably should. But that is the expectation.

src/cascade/model/grids.py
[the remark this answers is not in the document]
alec-deason: Interpolation (in prior space) is constant, not bilinear is what I mean to say.

src/cascade/model/grids.py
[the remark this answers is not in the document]
alec-deason: So in this interface ages are just ages and times are just times. It isn't until we translate to the Dismodfile where they get transformed into indexes.

src/cascade/model/compartmental.py
alec-deason: Oh, I forgot about raw strings. That is absolutely the right way to do this.

PR #16, opened 2018-08-07
feature/DISMODAT-50-load-asdr-to-t3

nrhawkins:


src/cascade/input_data/db/asdr.py
```diff
@@ -0,0 +1,153 @@
...
+    demo_dict = get_demographics(gbd_team="epi", gbd_round_id=GBD_ROUND_ID)
+    age_group_ids = demo_dict["age_group_id"]
+    sex_ids = demo_dict["sex_id"]
+
+    asdr = get_envelope(
+        location_id=parent_loc,
+        gbd_round_id=GBD_ROUND_ID,
+        age_group_id=age_group_ids,
+        sex_id=sex_ids,
+        with_hiv=1,
```
alec-deason: I know that their docs and code both treat `with_hiv` and `rates` as integers but that's absurd, let's not leak it into our code. Booleans will also work and we should use those.

PR #17, opened 2018-08-09
feature/DISMODAT-245-bash-scripts-singularity-epiviz

nrhawkins:
Moved the singularity and epiviz scripts over and into the scripts dir.

Is this the structure we want inside /scripts?

### DIFF
--- a/scripts/automation/dismod/build_singularity.sh
+++ b/scripts/automation/dismod/build_singularity.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+# Options:
+#   -i: Required, name of the image that you want to build, hosted on reg.ihme.uw.edu. For example, core/rstudio:3.4.3.1
+#   -f: Required, path to the file where you want to save your image
+#   -t: number of threads to use when pulling Dockerfile and building Singularity image. Default: 2. 
+#       Only applies on cluster-prod -- on cluster-dev, will use all CPUs assigned to your qlogin session.
+
+## Example:
+# sh /home/j/temp/grant/build_singularity.sh -i core/rstudio:3.4.3.1 -f /share/singularity-images/test_deploy/shell_test2.img -t 8
+
+# Set arguments for options
+while getopts "i::f::t::" opt; do
+  case $opt in
+    i ) image_name=$OPTARG
+    ;;
+    f ) file_path=$OPTARG
+    ;;
+    t ) threads=$OPTARG
+    ;;
+    \?) echo "Invalid option -$OPTARG" >&2
+    ;;
+  esac
+done
+
+
+# function to set environmental variables for build-time configuration
+# Mostly to constrain the number of python threads and to ensure that Singularity does not save cached images to a user's H drive.
+function set_env_variables {
+  echo "Setting environmental variables"
+  export SINGULARITY_CACHEDIR="/tmp"
+  export SINGULARITY_CACHE="/tmp"
+  export SINGULARITY_DISABLE_CACHE="true"
+  export SINGULARITY_PYTHREADS="$threads"
+}
+
+# function to submit a new job
+function build_singularity {
+  if [ -z $threads ]
+    then
+      threads=2
+  fi 
+
+  if [ -z $image_name ]
+    then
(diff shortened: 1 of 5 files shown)


scripts/automation/dismod/create_singularity.sh
```diff
@@ -0,0 +1,15 @@
...
+#$ -pe multi_slot 5
+#$ -cwd
+#$ -j y
+#$ -o /share/temp/sgeoutput/adolgert/output.txt
+#$ -N SingDismod
+#$ -S /bin/bash
+
+set -x
+THREADS=5
+IMAGE=dismod_at:0.0.1
```
adolgert: This part has 0.0.1 built-in. It's a key place to automate. Not for this PR.

scripts/automation/dismod/build_singularity.sh
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env bash
```
adolgert: This script is a hacked copy of the build_singularity.sh that Grant made. He may have updated his since this was copied. I can't recall why I hacked it, either. In short, we need to go back and try Grant's latest version, using his instructions on the hub.

scripts/dbg_full.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# This script is for debugging. The run_cascade.sh will call it
+# so that we can see whether run_cascade.sh is working.
+
+OUTDIR="/share/temp/sgeoutput/$USER/errors"
```
adolgert: This hasn't changed.

scripts/run_full.sh
```diff
@@ -0,0 +1,34 @@
...
+    exit 4
+fi
+
+if [ $# -gt 1 ]; then
+    DEBUG=$2
+else
+    DEBUG=0
+fi
+
+if test -r ~/EPIAT_DIR
```
adolgert: This is new to me. It's a way to use your own environment. It uses the existence of a directory instead of using a configuration file in ~/.config/. Somehow I think the configuration file would be better? But we should discuss.

scripts/automation/dismod/create_singularity.sh
alec-deason: :+1: :confetti_ball:

scripts/run_full.sh
nrhawkins: I think you're the author of line 15?  

the next line is: EPI_DIR="$(cat ~/EPIAT_DIR)"

so, it appears that EPIAT_DIR would be a readable file that would contain the value for EPI_DIR?

scripts/automation/dismod/build_singularity.sh
nrhawkins: I have a copy of Grant's latest version.  There are some lines that show up in the diff, that look the same to me; for example, one example is:
41c41
<   fi
---
>   fi

There are also two places where I can see differences in the text:

1)  In our copy of build_singularity.sh, "dev" below is "devnotgonnahappen", is there a good reason for that?

## Set CPUs to pin to using numactl -- for dev cluster, CPU enforcement is enabled so we can pull the CPUs directly. For cluster-prod, need
 to generate a list
    if [ $SGE_CLUSTER_NAME == "dev" ]
      then
        cpu_list="$(cat /proc/$$/status | grep '^Cpus_allowed_list:' | awk '{print $NF}')"
      else
		target_thread=$(($threads - 1))
        cpu_list="0-$target_thread"
      fi

2) In our copy of build_singularity.sh,  we write "singularity" in the line below,

numactl -C +$cpu_list singularity build $file_path docker://reg.ihme.uw.edu/$image_name

as: 
/share/local/singularity-2.4.2/bin/singularity

is there a good reason for that?

adolgert: I think the change to just "singularity" means the path is updated for everybody. The part about devnotgonnahappen is that the allowed CPUS list is a good idea but not good in practice, so I don't want to do it. You are pinning processes to CPUs, but other processes won't play along, so it won't work.

scripts/run_full.sh
adolgert: I suggest we could make a dotfile that redirects user code to use a particular install. This works meanwhile.

PR #22, opened 2018-08-15
Make index columns visible in the dismodfile interface

alec-deason:
This is for Greg

### DIFF
--- a/src/cascade/dismod/db/wrapper.py
+++ b/src/cascade/dismod/db/wrapper.py
@@ -60,7 +60,7 @@ def _validate_data(table_definition, data):
                 # Permit np.float because an int column with a None is cast to float.
                 # Same for object. This is cast on write.
                 # Because we use metadata, this will be converted for us to int when it is written.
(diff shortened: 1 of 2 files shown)


src/cascade/dismod/db/wrapper.py
```diff
@@ -73,8 +73,8 @@ def _validate_data(table_definition, data):
             elif expected_type is str:
                 if len(data) > 0:
                     # Use iloc to get the first entry, even if the index doesn't have 0.
-                    actual_type = type(data[column_name].iloc[0])
-                    correct = np.issubdtype(actual_type, np.str_) or actual_type == type(None)
+                    entry = data[column_name].iloc[0]
+                    correct = np.issubdtype(type(entry), np.str_) or entry is None
```
adolgert: thank you for fixing this! I should not have left that.

alec-deason: There's another weird one in there where I just disabled the linter. I couldn't figure out how to do it more cleanly.

PR #25, opened 2018-08-17
Backend for smoothing API and other internal model -> dismodfile translation

alec-deason:
This adds a backend for the API in https://github.com/ihmeuw/cascade/pull/15 and integrates it with a lightly cleaned up version of Drew's Dismodfile builder. There is a lot of work to be done on making that part of the system clear and robust, but this starts it.


alec-deason:
This also adds some new things to the ModelContext to better support the internal model. That's another place that needs more thought than this gives it.

alec-deason:
Drew points out that my pretty printer has filled this PR with noise, for which I apologize. I'm going to figure out a way not to do that in future.

docs/refmanual.rst
```diff
@@ -8,4 +8,5 @@ Cascade-AT Reference Manual
    core
    dismod
    executor
+   model
```
nrhawkins: this almost looks alphabetical, just saying.

src/cascade/executor/no_covariate_main.py
```diff
@@ -158,95 +169,57 @@ def age_year_from_data(df):
...
-
-
-def internal_model(model_context, inputs):
-    config = model_context.parameters
-    model = model_context.input_data
+    ages = constraint["age_start"].unique().tolist()
+    times = constraint["year_start"].unique().tolist()
+    grid = AgeTimeGrid(ages, times)
+    smoothing_prior = PriorGrid(grid)
+    smoothing_prior[:, :].prior = UniformPrior(-np.inf, np.inf, 0)
```
adolgert: This prior with the infinities... does it work for Dismod? It's a completely uninformative prior. That's what the are called. Difficult to use.

src/cascade/executor/no_covariate_main.py
```diff
@@ -158,95 +169,57 @@ def age_year_from_data(df):
...
+
+    value_prior = PriorGrid(grid)
+    # TODO: change the PriorGrid API to handle this elegantly
+    for _, row in constraint.iterrows():
+        value_prior[row["age_start"], row["year_start"]].prior = ConstantPrior(row["mean"])
```
adolgert: Well, I like that it's clearly a constant prior.

src/cascade/executor/no_covariate_main.py
```diff
@@ -158,95 +169,57 @@ def age_year_from_data(df):
...
-        "omega": build_constraint(model.constraints),
-    }
-    LOGGER.debug(f"Omega constraint {model.smoothers['omega']}")
-    return model_context
+    for rate in config.non_zero_rates.split():
+        getattr(model.rates, rate).parent_smooth = default_smoothing
+
+    for rate in config.non_zero_rates.split() + ["prevalence"]:
+        integrand = getattr(model.outputs.integrands, RATE_TO_INTEGRAND[rate].name)
+        integrand.active = True
```
adolgert: this "active" flag.. seems like it could go wrong? Is there a way we can know what's active? Is this so we know what to write to the file, which integrands to include in a list?

src/cascade/executor/no_covariate_main.py
```diff
@@ -158,95 +169,57 @@ def age_year_from_data(df):
...
+        integrand.age_lower = min(default_smoothing.grid.ages)
+        integrand.age_upper = max(default_smoothing.grid.ages)
```
adolgert: I'm not sure these lines are right. Seems to be setting a min/max age for the output integrand.  Ohhh... this is for avgint table. Got it. That documentation is unclear, and I think this code isn't correct. The min/max isn't for the lowest possible. Each entry in the avgint has its own min and max, like 25-30 yr-olds.

src/cascade/executor/no_covariate_main.py
```diff
@@ -255,28 +228,30 @@ def construct_database():
     model_context.parameters.non_zero_rates = "iota rho chi omega"
 
     # Get the bundle and process it.
-    raw_inputs = data_from_csv(Path("measure.csv"))
+    raw_inputs = data_from_csv(Path(input_path))
```
adolgert: :)

src/cascade/model/compartmental.py
```diff
@@ -93,7 +93,7 @@ def lx(t):
 
 .. math::
 
-    P = \frac{C}{S+C}
+    P = \\frac{C}{S+C}
```
adolgert: When you see this, turn it the comment into raw strings. It's my fault for the error. I think the raw strings make the error very much go away with one character.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,235 @@
+"""Support for dismod concepts that are expressed over grids: age-time, smoothings and value priors.
+"""
+from math import isclose, isinf
+
+import numpy as np
+
+GRID_SNAP_DISTANCE = 1 / 365
```
adolgert: snap seems too large. Maybe 1/100 of this? b/c days matter in first age group, and 36500 inverted is still way above floating point differentiation.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,235 @@
...
+
+        Args:
+            age_start: the lowest age in the grid
+            age_end: the highest age in the grid
+            age_step: the spacing of age points
+            time_start: the earliest time in the grid
+            time_end: the latest time in the grid
+            time_step: the spacing of time points
+        """
+        ages = np.arange(age_start, age_end, age_step)
```
adolgert: I have a slight preference for linspace here. Let me sell you on that: Both linspace and arange require that the user be careful about matching step sizes to beginning and end. With linspace, you are guaranteed that the end is included in the range, but with arange you are guaranteed that the step sizes are right. I think it's more important to go to age 100 than to have exactly 2-year intervals along the way. Hmmm. I don't know. both are annoying.

src/cascade/dismod/db/wrapper.py
```diff
@@ -267,10 +279,9 @@ def _check_column_types_actually_written(self):
             for table_name, table_definition in self._table_definitions.items():
                 introspect = text(f"PRAGMA table_info([{table_name}]);")
                 results = connection.execute(introspect)
-                if results.returns_rows:
-                    table_info = results.fetchall()
-                else:
-                    continue  # Not all tables are in all databases.
+                if not results.returns_rows:
+                    continue
+                table_info = results.fetchall()
```
nrhawkins: I have heard that Code Complete likes if/else.  
I think you're saying, no else is ok.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,235 @@
...
+        times = np.arange(time_start, time_end, time_step)
+
+        return cls(ages, times)
+
+    def __init__(self, ages, times, snap_distance=GRID_SNAP_DISTANCE):
+        self._ages = tuple(sorted(set(ages)))
```
adolgert: This set should take snap distance into account. That's a complicated maneuver. Maybe that's a later feature. For instance, if two values are within 1/365 of each other, and you have to set a value, which of the two do you choose? What if one value is near two others, but those two are further than 1/365 away from the first? Do we snap all values to the closest that's 1/365? Meaning multiply by 365, round them, convert to int, and then make the set?

src/cascade/model/grids.py
```diff
@@ -0,0 +1,235 @@
...
+
+    Examples:
+        >>> grid = AgeTimeGrid.uniform(age_start=0,age_end=120,age_step=5,time_start=1990,time_end=2018,time_step=1)
+        >>> d_time = PriorGrid(grid)
+        >>> #Set a prior for the whole grid:
+        >>> d_time[:, :].prior = GaussianPrior(0, 0.1)
+        >>> #Set a prior for a band of ages
+        >>> d_time[0:15, :].prior = GaussianPrior(1, 0.01)
+        >>> #Or a single year
+        >>> d_time[:, 1995].prior = GaussianPrior(0, 3)
```
adolgert: I like this kind of demo.

src/cascade/model/grids.py
```diff
@@ -0,0 +1,235 @@
...
+    def _push_prior(self, lower_age, upper_age, lower_time, upper_time, prior):
+        """Push a new region and prior combination onto the stack. When
+        querying, more recent priors take precedence.
+        """
+        _validate_region(self.grid, lower_age, upper_age, lower_time, upper_time)
+        lower_age = self.grid.to_age(lower_age)
+        upper_age = self.grid.to_age(upper_age)
+        lower_time = self.grid.to_time(lower_time)
+        upper_time = self.grid.to_time(upper_time)
+        self._priors.append(((lower_age, upper_age, lower_time, upper_time), prior))
```
adolgert: So there is a stack. I was wondering about this. We could apply it immediately. Why keep the stack? Just asking why.

src/cascade/dismod/serialize.py
```diff
@@ -2,25 +2,103 @@
...
+    return bundle_fit
+
+
 def enum_to_dataframe(enum_name):
     """Given an enum, return a dataframe with two columns, name and value."""
     return pd.DataFrame.from_records(
         np.array(
             [(measure, enum_value.value) for (measure, enum_value) in enum_name.__members__.items()],
-            dtype=np.dtype([('name', object), ('value', np.int)])
+            dtype=np.dtype([("name", object), ("value", np.int)]),
```
nrhawkins: I have noticed that you're saying that lists should end with a comma

docs/refmanual.rst
alec-deason: Well, my mother was a librarian.

src/cascade/dismod/db/wrapper.py
alec-deason: What would the else add when there's an explicit exit in the if? Code Complete seems to be confused about a lot of things.

src/cascade/dismod/serialize.py
alec-deason: Yes. I put them most places, the pretty printer puts them _everywhere_. Trailing commas make it very fast to add stuff to the list or move items around. It also makes diffs shorter for simple changes in the list.

tests/model/test_grids.py
```diff
@@ -0,0 +1,85 @@
...
+from cascade.model.priors import GaussianPrior
+
+
+@pytest.fixture
+def age_time_grid():
+    return AgeTimeGrid.uniform(age_start=0, age_end=120, age_step=5, time_start=1990, time_end=2018, time_step=1)
+
+
+# def test_PriorGrid__development_target():
+#    # Make a grid using a standard structure. Other options might be
```
adolgert: These commented-out parts. Should they be comments or code or gone?

examples/theory_predict.py
```diff
@@ -31,8 +31,11 @@
 from scipy.stats import gamma
 
 from cascade.model import (
-    siler_default, prevalence_solution, integrands_from_function, omega_from_mu,
```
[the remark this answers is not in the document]
alec-deason: Yeah, I'm sorry. :(

src/cascade/core/context.py
```diff
@@ -1,5 +1,6 @@
-from cascade.core.parameters import ParameterProperty
-from cascade.core.input_data import InputData
+from .parameters import ParameterProperty
+from .input_data import InputData
+from cascade.model.rates import Rate
```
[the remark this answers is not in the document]
alec-deason: Right.

src/cascade/core/context.py
```diff
@@ -17,6 +18,63 @@ class _ModelParameters:
...
+    __slots__ = ["integrands"]
+
+    def __init__(self):
+        self.integrands = _Integrands()
+
+
+class _Integrand:
+    def __init__(self, name):
+        self.name = name
+        self.active = False
```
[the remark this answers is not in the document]
alec-deason: I don't know what makes it active so I am requiring that you explicitly set the flag. If you have a better way, I'm open.

src/cascade/dismod/db/metadata.py
```diff
@@ -37,7 +38,7 @@ def _big_integer_callback(element, compiler, **kw):
     return "integer primary key"
 
 
-@compiles(Float, 'sqlite')
+@compiles(Float, "sqlite")
```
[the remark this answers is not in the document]
alec-deason: I love single quotes but have finally been convinced that they're the wrong thing.

src/cascade/model/grids.py
alec-deason: I'm fine with either. This is a tool I made up based on my assumptions of what we might want. It looks like in practice this won't be very useful as the modelers will usually be explicitly setting up their grids. So I might argue for pulling this method completely.

src/cascade/model/priors.py
```diff
@@ -0,0 +1,178 @@
...
+        raise ValueError("Nu must be positive")
+
+
+class UniformPrior(_Prior):
+    density = "uniform"
+
+    def __init__(self, lower, upper, mean=None, name=None):
+        super().__init__(name=name)
+        if mean is None:
+            mean = (upper - lower) / 2 + lower
```
nrhawkins: this is the same as (upper + lower) / 2 , just wondering about it

src/cascade/model/priors.py
```diff
@@ -0,0 +1,178 @@
...
+        }
+
+
+class LogGaussianPrior(_Prior):
+    density = "log_gaussian"
+
+    def __init__(self, mean, standard_deviation, eta, lower=float("-inf"), upper=float("inf"), name=None):
+        super().__init__(name=name)
+        _validate_bounds(lower, mean, upper)
+        _validate_standard_deviation(standard_deviation)
```
nrhawkins: I like seeing all these "validates".  It helps me; i.e. because it's not exactly textbook, i.e. not taught.

src/cascade/model/grids.py
alec-deason: Good point. Maybe the right thing to do is reduce the snap distance down to floating point error and call it good? That doesn't actually resolve your question but it makes it unlikely to be an issue in practice.

src/cascade/model/rates.py
```diff
@@ -0,0 +1,76 @@
...
+
+        self.value_priors = value_priors
+        self.d_age_priors = d_age_priors
+        self.d_time_priors = d_time_priors
+
+    def _validate_grids(self, priors):
+        grids = [ps.grid for ps in [self.value_priors, self.d_age_priors, self.d_time_priors, priors] if ps]
+        if grids:
+            if not all([grids[0] == g for g in grids]):
+                raise ValueError("Smooth cannot contain priors on hetrogenious grids")
```
nrhawkins: ok, I like the idea of "genius" grids.

maybe this should be: heterogeneous ?

src/cascade/model/grids.py
alec-deason: Unless I store it as a value on each grid point (which would be a reasonable thing to do) it's simpler not to resolve it immediately because that would require me to do rectangle overlap operations. Now that I understand how the priors are used better storing them on the points actually makes more sense than I thought it did initially.

tests/model/test_grids.py
alec-deason: Crap, no that's a mistake. I wanted to leave that in but I also wanted to see coverage with it gone so that I remembered to write explicit tests for each of the pieces. It hides coverage by exercising everything without individually testing anything.

src/cascade/model/priors.py
alec-deason: It is. The other way made more sense to me at the time but now it just seems confusing. Yours is more clear.

src/cascade/model/priors.py
alec-deason: Validation at API boundaries (which this is) is always a good idea.

src/cascade/model/rates.py
alec-deason: Yesyes

src/cascade/core/context.py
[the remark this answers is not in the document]
adolgert: Where does the active flag get used? It's not for the integrands table. That's just a list of integrands. Maybe the active flag is never used?

src/cascade/model/grids.py
adolgert: I suspect we will have things that generate age ranges but that they will be more complicated because the young ages will have finer grid points and the oldest more granular regions. Removing this from the class is kinda fine, I suppose.

src/cascade/model/grids.py
adolgert: Yes, wait. Is it a problem if two ages are nearly the same but not quite? .... I don't think it is. Nice!

src/cascade/model/grids.py
adolgert: The layers is cool. It doesn't show up in the API, though. Putting them on the points would work.

src/cascade/model/priors.py
adolgert: This is also nicer than the validations in Alex's code b/c they are validated in logically-related places, as opposed to a long list. Less brittle this way.

src/cascade/core/context.py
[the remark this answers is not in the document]
alec-deason: src/cascade/executor/no_covariate_main.py:211

It's used when the setup code interprets the user's non_zero_rates list. It's more than likely that I'm approaching this wrong, especially since you seem to find it unclear.

src/cascade/dismod/db/metadata.py
```diff
@@ -20,6 +20,7 @@
 # The method is here:
 #   http://docs.sqlalchemy.org/en/latest/core/compiler.html?highlight=compiler#module-sqlalchemy.ext.compiler
 
+
```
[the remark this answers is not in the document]
alec-deason: Such is life

src/cascade/dismod/db/wrapper.py
```diff
@@ -161,8 +161,10 @@ def __init__(self, engine, avgint_columns, data_columns):
         self._table_definitions = self._metadata.tables
         self._table_data = {}
         self._table_hash = {}
-        add_columns_to_avgint_table(self._metadata, avgint_columns)
-        add_columns_to_data_table(self._metadata, data_columns)
+        if avgint_columns:
```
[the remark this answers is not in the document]
alec-deason: I'm avoiding having mutable default arguments on principle, though it actually wouldn't be a problem here unless someone changed the implementation. Though it would be a problem if they changed the implementation... Also the linter will yell at me if I do it. Hence the conditionals.

src/cascade/dismod/db/wrapper.py
```diff
@@ -178,14 +180,22 @@ def make_densities(self):
         so this puts them all in.
         """
         self.density = pd.DataFrame({"density_name": [x.name for x in DensityEnum]})
+        self.density["density_id"] = self.density.index
+
+    def __dir__(self):
+        attributes = list(self._table_definitions.keys())
+        attributes.extend(super().__dir__())
+        return attributes
```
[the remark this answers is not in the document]
alec-deason: I have been seduced into the world of tab completion.

src/cascade/executor/no_covariate_main.py
```diff
@@ -51,19 +55,15 @@
 def cached_bundle_load(context, bundle_id, tier_idx):
     cache_bundle = Path(f"{bundle_id}.pkl")
     if cache_bundle.exists():
-        LOGGER.info(f"Reading bundle from {cache_bundle}. "
-                    f"If you want to get a fresh copy, delete this file.")
+        LOGGER.info(f"Reading bundle from {cache_bundle}. " f"If you want to get a fresh copy, delete this file.")
```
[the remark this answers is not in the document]
alec-deason: This is the auto formatter and that's super ugly, I agree. I like this tool but it's certainly not perfect. I'll clean this up.

src/cascade/executor/no_covariate_main.py
```diff
@@ -158,95 +169,57 @@ def age_year_from_data(df):
...
-        "const_value": constraint["mean"],
-    })
-
-
-def internal_model(model_context, inputs):
-    config = model_context.parameters
-    model = model_context.input_data
+    ages = constraint["age_start"].unique().tolist()
+    times = constraint["year_start"].unique().tolist()
+    grid = AgeTimeGrid(ages, times)
```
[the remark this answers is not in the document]
alec-deason: It actually doesn't even need uniqued values at this layer. The constructor takes care of both sorting and uniquing. I'm not sure why I wrote this bit this way, it's redundant.

src/cascade/executor/no_covariate_main.py
alec-deason: This is functionally identical to what you had in your code, unless I misunderstood what you were doing which is possible. Whether it makes sense or not, I defer to you.

src/cascade/executor/no_covariate_main.py
alec-deason: It renders as a uniform with lower=upper=mean, which is what Brad does internally anyway when you set const_value.

src/cascade/executor/no_covariate_main.py
alec-deason: It's so we know what to write into the integrand table. I'm not totally happy with this either and am open to suggestions. What I want is:

1. Ease of access to the integrands by name without complicated setup on the client side
2. Explicit ordering when you iterate
3. Explicit indication of which integrands are included in the model

This is the cleanest way I could come up with to do all that.

src/cascade/executor/no_covariate_main.py
alec-deason: Oh... Ok. I'll clean this up.

src/cascade/model/grids.py
alec-deason: I'm going to push it down to epsilon.

src/cascade/model/grids.py
alec-deason: I'm going to keep it and switch to linspace. Even if the modelers don't use it, it's nice for examples.

src/cascade/model/grids.py
alec-deason: I'm realizing that if you have grids where the points aren't at integer values or simple floats and the snap distance is small then this gets awkward to work with. Maybe that won't matter in practice.

src/cascade/model/grids.py
alec-deason: Ok. After realizing the difficulty of having non-integer/simple float points in the grids I'm going back to arange.

src/cascade/model/grids.py
alec-deason: Yeah, it's just an implementation detail.

alec-deason:
I believe I have now responded to all the feedback on this PR. If everybody's OK with it then I'm going to go ahead and merge this along with the one for the frontend.

alec-deason:
Wait, no. There's another comment that I haven't responded to but which github made vanish.

alec-deason:
Ok. Now.

PR #26, opened 2018-08-17
Feature/dismodat 221 full save results at

nrhawkins:
This "save_results" uses the Central Comp save_results_at() which has model_version_id as a parameter.

Added a scratch_maker to the context.py
Added a property, dismodfile, to the ExecutionContext

save_results_at() has this list of parameters: 

def save_results_at(input_dir,
                    input_file_pattern,
                    modelable_entity_id,
                    description,
                    measure_id,
                    model_version_id=None,
                    year_id=[1990, 1995, 2000, 2005, 2010, gbd.GBD_ROUND],
                    sex_id=[1, 2],
                    metric_id=3,
                    db_env='prod',
                    gbd_round_id=gbd.GBD_ROUND_ID)

In the review, you can compare this with those used in: cascade.saver.save_model_results which calls this function.


src/cascade/saver/generate_draws.py
```diff
@@ -0,0 +1,104 @@
...
+    CODELOG.debug(f"draws shape: {draws.shape}")
+    CODELOG.debug(f"draws cols: {draws.columns}")
+
+    draw_df = avgint.merge(draws, how="left", on="avgint_id").drop(
+        columns=["avgint_id", "integrand_id"])
+
+    CODELOG.debug(f"draw_df shape (expect {num_draws+5} cols):{draw_df.shape}")
+    CODELOG.debug(f"draw_df row 1 {draw_df.iloc[0]}")
+
+    if not draw_df.shape[0]:
```
alec-deason: This seems like something you should catch earlier or not check. It's effectively a (weak) check for the precondition that avgint_ids match in the two tables. I'm not sure we really need that check but if you're going to do it you should do the strong version and at the top of the function. So rather than check that _some_ avgint_ids match, check that they _all_ match.

src/cascade/saver/save_model_results.py
```diff
@@ -0,0 +1,92 @@
...
+        (str) of the mvid returned by save_results
+
+    """
+
+    dm_file = execution_context.dismodfile
+
+    if not dm_file:
+        raise ValueError("Must provide a dismod file")
+
+    draws_df = generate_draws_table(dm_file)
```
alec-deason: I'd like to keep the pattern where we drive database access code from outside. So you would have a method that takes a context, calls `generate_draws_table` and then passes the resulting table into this function. Does that make sense? Code that has direct database access stuff in it is always structured as utility functions and the sequences that drive the programs behavior are in higher level functions.

src/cascade/saver/save_model_results.py
```diff
@@ -0,0 +1,92 @@
...
+
+    CODELOG.debug(draws_df.columns)
+    CODELOG.debug(draws_df.iloc[0])
+
+    with execution_context.scratch_dir() as scratch_dir:
```
alec-deason: Using this as a context manager is perfect. But you should try to make it so only code that needs the scratch_dir is inside the manager, move everything else above or below. That way it's easy to see exactly what the context manager effects.

src/cascade/saver/save_model_results.py
```diff
@@ -0,0 +1,92 @@
...
+        CODELOG.debug(scratch_dir)
+        file_path = (scratch_dir / DRAWS_INPUT_FILE_PATTERN)
+
+        model_version_id = execution_context.parameters.model_version_id
+        modelable_entity_id = execution_context.parameters.modelable_entity_id
+
+        CODELOG.debug(f"modelable_entity_id {modelable_entity_id}")
+
+        if not modelable_entity_id:
```
alec-deason: Keep preconditions checks at the top of the function whenever that's practical. But this is an internal function, you shouldn't need to check this at all. This is something code at this layer should trust it's caller to get right, especially since the cost of getting it wrong is just a slightly more difficult error message a little further along.

tests/saver/test_generate_draws.py
```diff
@@ -0,0 +1,165 @@
...
+
+from cascade.core.context import ExecutionContext
+import cascade.saver
+from cascade.saver.generate_draws import (
+    generate_draws_table,
+    pure_generate_draws,
+    retrieve_prediction_tables)
+from cascade.dismod.db.wrapper import _get_engine, DismodFile
+
+# DRAWS_DB_FILE = "save_results.db"
```
alec-deason: If it's unused delete it.

tests/saver/test_generate_draws.py
```diff
@@ -0,0 +1,165 @@
...
+                "db_env": "dev"}
+    execution_context = ExecutionContext()
+    execution_context.parameters = defaults
+    execution_context._dismodfile = dismod_file
+
+    return execution_context
+
+
+@pytest.fixture(scope="module")
+def predict_df_ragged(predict_df):
```
alec-deason: Nice, good case to check.

tests/saver/test_generate_draws.py
```diff
@@ -0,0 +1,165 @@
...
+    return avgint_df(), predict_df()
+
+
+def get_draws(avgint, predict):
+    return expected_draws_df()
+
+
+@pytest.fixture
+def fake_generate_draws(monkeypatch):
+    monkeypatch.setattr(cascade.saver.generate_draws,
```
alec-deason: Why mock this when you can just build the DismodFile directly?

tests/saver/test_save_model_results.py
```diff
@@ -0,0 +1,201 @@
...
+from cascade.core.context import ExecutionContext
+import cascade.saver
+import cascade.saver.save_model_results as smr
+from cascade.dismod.db.wrapper import _get_engine, DismodFile
+
+DRAWS_INPUT_FILE_PATTERN = "all_draws.h5"
+
+
+@pytest.fixture(scope="module")
+def predict_df():
```
alec-deason: You have a fixture for this in both test files. You can move this into a module level conftest.py if it makes sense for them to share.

src/cascade/saver/save_model_results.py
```diff
@@ -0,0 +1,92 @@
...
+            model_title,
+            measures_to_save,
+            model_version_id,
+            db_env=db_env)
+
+        CODELOG.debug(f"mvid_df: {model_version_id_df.iloc[0, 0]}")
+
+        # context.scratch_dir() handles clean-up
+
+        return str(model_version_id_df.iloc[0, 0])
```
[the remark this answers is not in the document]
adolgert: That one is on me. It isn't always an integer. I would argue that, since it isn't use for addition, it never was an integer. But also there was talk of using 1234.7 with the dot at the end for some kind of internal versioning.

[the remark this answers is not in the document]
alec-deason: In the underlying database it's an `int` but if people are talking about changing that then that's fine. But if they do I guarantee half the system in the building will burn to the ground. Also, it's a primary key that's frequently used in complex joins. Having that be a non-int type sounds terrible. So either I misunderstand something or someone else does. Either way, we can treat it as a string with some kind of semantic versioning inside our system, that works just fine.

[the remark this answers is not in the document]
alec-deason: Actually, no. Making this a string in our system has weird implications because we must be able to query the external dbs using this value and there it is an int. I think we should keep it as an int until CC changes it to a string.

src/cascade/saver/generate_draws.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+import logging
+
+import pandas as pd
+
+CODELOG = logging.getLogger(__name__)
+MATHLOG = logging.getLogger(__name__)
+
+INTEGRAND_ID_TO_MEASURE_ID_DF = pd.DataFrame([
```
alec-deason: This belongs with the db code. Converting back and forth between our internal representation (integrand_id) and the external representation (measure_id) is the sort of normalization which should happen with the database code so that all the logic directly related to the external dbs lives in one place.

PR #28, opened 2018-08-22
feature/DISMODAT-289-update-install-doc

nrhawkins:
added path to our miniconda3 for virtualenv call
added call for setting the print_level_fixed option (so, say, Liane, could get more info about a "not converging?" job)
fixed typos

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -6,10 +6,12 @@ Installation of Cascade
 Cascade interacts with Dismod-AT underneath. Cascade runs Dismod-AT within
 the IHME infrastructure. Clone it from
 `Cascade on Github <https://github.com/ihmeuw/cascade>`_.
-We recommend you create a virtual environment into which to install
-the code. If you have Python3, virtualenv is part of it::
 
-    virtualenv ./env_path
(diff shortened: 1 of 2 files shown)


docs/install.rst
```diff
@@ -6,10 +6,12 @@ Installation of Cascade
...
 the IHME infrastructure. Clone it from
 `Cascade on Github <https://github.com/ihmeuw/cascade>`_.
+
 We recommend you create a virtual environment into which to install
-the code. If you have Python3, virtualenv is part of it::
+the code. You need python3 to run virtualenv.  You can use our miniconda3 
+installed on the cluster::
 
-    virtualenv ./env_path
+    /ihme/code/dismod_at/pyenv/versions/miniconda3-4.5.1/bin/virtualenv ./env_path
```
alec-deason: If we're going to maintain a version of virtualenv ourselves (and I don't see that there is one maintained by central comp so we may have to) then let's put it in a place with fewer version numbers, say `/ihme/code/dismod_at/pyenv/versions/miniconda`. That can be a symlink to the current version with all the version numbers if that makes sense.

PR #29, opened 2018-08-23
Add command line argument for non-zero-rates

alec-deason:


alec-deason:
How does this look for docs?

PR #30, opened 2018-08-24
Make input data optional

alec-deason:
This is in support of Greg's use case where he wants to be able to construct constraints either by hand or by running custom analysis code on inputs without having to put those inputs into the data table.

### DIFF
--- a/src/cascade/dismod/serialize.py
+++ b/src/cascade/dismod/serialize.py
@@ -2,6 +2,7 @@
 Converts the internal representation to a Dismod File.
 """
 import logging
+import warnings
 import time
 
 import numpy as np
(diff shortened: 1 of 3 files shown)


src/cascade/dismod/serialize.py
```diff
@@ -128,18 +129,46 @@ def make_log_table():
...
-    constraints = observations_to_data(context.input_data.constraints, hold_out=1)
-    total_data = pd.concat([observations, constraints], ignore_index=True)
-    # Why a unique string name?
-    total_data["data_name"] = total_data.index.astype(str)
+    total_data = []
+    if context.input_data.observations is not None:
+        total_data.append(observations_to_data(context.input_data.observations))
+    if context.input_data.constraints is not None:
+        total_data.append(observations_to_data(context.input_data.constraints, hold_out=1))
+
```
adolgert: This looks like the central section of the PR? The idea is that there isn't any input data but there can be observations and constraints. OK.

PR #36, opened 2018-09-05
Feature/dismodat 273 load country covariates

nrhawkins:


src/cascade/input_data/db/ccov.py
```diff
@@ -0,0 +1,57 @@
...
+            raise ImportError(f"Required package db_queries not found")
+
+    get_covariate_estimates = DummyGetCovariateEstimates()
+
+
+CODELOG = logging.getLogger(__name__)
+MATHLOG = logging.getLogger(__name__)
+
+
+def country_covariates(execution_context, demographics):
```
alec-deason: I think it's going to make more sense to structure this such that it retrieves a single country covariate. Based on how the GUI is setup I think we will end up wanting an outer loop that goes through the covariates one at a time and attaches them to the model.

PR #37, opened 2018-09-06
Bugfix/allow writing of covariate columns

alec-deason:
This lets us define covariate columns from the actual schema of pandas tables. It also makes the precondition checks on the column names tighter.

One limitation with this is that it won't remove columns, only add. I'm not sure if that's an issue.

### DIFF
--- a/src/cascade/dismod/db/metadata.py
+++ b/src/cascade/dismod/db/metadata.py
@@ -4,6 +4,8 @@ This describes the tables in the sqlite file that Dismod reads.
 import enum
 import logging
 
+import numpy as np
+
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy import Column, Integer, String, Float, Enum, ForeignKey
 from sqlalchemy import BigInteger
@@ -532,7 +534,7 @@ class Var(Base):
     covariate_id = Column(Integer(), nullable=False)
 
 
-_TYPE_MAP = {str: String, int: Integer, float: Float}
+_TYPE_MAP = {str: String, int: Integer, float: Float, np.dtype("int64"): Integer, np.dtype("float64"): Float}
 
 
 def add_columns_to_avgint_table(metadata, column_identifiers):
--- a/src/cascade/dismod/db/wrapper.py
+++ b/src/cascade/dismod/db/wrapper.py
@@ -183,16 +183,16 @@ class DismodFile:
 
     def update_table_columns(self, table_definition, table):
         new_columns = table.columns.difference(table_definition.c.keys())
-        new_columns = {c: table.dtypes[c] for c in new_columns}
+        new_column_types = {c: table.dtypes[c] for c in new_columns}
 
-        bad_column_names = [c for c in new_columns.keys() if not c.startswith("x_")]
+        bad_column_names = [c for c in new_columns if not c.startswith("x_")]
         if bad_column_names:
             raise ValueError(f"Covariate column names must start with 'x_'. Malformed names: {bad_column_names}")
 
         if table_definition.name == "avgint":
-            add_columns_to_avgint_table(self._metadata, new_columns)
+            add_columns_to_avgint_table(self._metadata, new_column_types)
         elif table_definition.name == "data":
-            add_columns_to_data_table(self._metadata, new_columns)
+            add_columns_to_data_table(self._metadata, new_column_types)
         else:
             raise ValueError(f"Can't add columns to {table.name}")
 
--- a/src/cascade/dismod/db/wrapper.py
+++ b/src/cascade/dismod/db/wrapper.py
@@ -146,25 +146,19 @@ class DismodFile:
     column name to column type.
     """
 
-    def __init__(self, engine=None, avgint_columns=None, data_columns=None):
+    def __init__(self, engine=None):
         """
         The columns arguments add columns to the avgint and data
         tables.
 
         Args:
             engine: A sqlalchemy engine
-            avgint_columns (dict): From columns to types.
-            data_columns (dict): From columns to types.
         """
         self.engine = engine
         self._metadata = deepcopy(Base.metadata)
         self._table_definitions = self._metadata.tables
         self._table_data = {}
         self._table_hash = {}
-        if avgint_columns:
-            add_columns_to_avgint_table(self._metadata, avgint_columns)
-        if data_columns:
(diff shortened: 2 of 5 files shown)


tests/dismod/db/test_wrapper.py
```diff
@@ -189,42 +189,85 @@ def test_validate_data__missing_column():
...
+    base_file.data = new_data
+    base_file.flush()
+
+    new_dm = DismodFile(base_file.engine)
+    assert list(new_dm.data.x_s_source) == [0]
+    assert list(new_dm.data.x_sex) == [2.0]
+
+
+def test_write_covariate_column__bad_name(base_file):
+    new_data = pd.DataFrame(
```
nrhawkins: Looks like this one is the same as above.  Maybe share it?

alec-deason: Valid

src/cascade/dismod/db/wrapper.py
```diff
@@ -187,15 +181,40 @@ def __dir__(self):
...
             table = self._table_definitions[table_name]
-            with self.engine.connect() as conn:
-                data = pd.read_sql_table(table.name, conn)
+            try:
+                with self.engine.connect() as conn:
+                    data = pd.read_sql_table(table.name, conn)
+            except ValueError as e:
+                if str(e) != f"Table {table.name} not found":
+                    raise
+                data = pd.DataFrame({k: pd.Series(dtype=v.type.python_type) for k, v in table.c.items()})
```
adolgert: There is a function in this class, called `empty_table` that doesn't have the same list of column types, and it should.

tests/dismod/db/test_wrapper.py
```diff
@@ -189,42 +189,69 @@ def test_validate_data__missing_column():
     assert "nonnullable_column" in str(excinfo.value)
 
 
-def test_two_files_different_columns():
```
adolgert: This test should have had a docstring. I wrote it. It's the only test that creates two DismodFile objects with different added columns, and we should still be able to do that with the code you've written, and still test that.

tests/dismod/test_serialize.py
```diff
@@ -234,7 +234,7 @@ def test_make_avgint_table(base_context):
 
 
 def test_make_rate_table(base_context):
-    dm = DismodFile(None, {}, {})
+    dm = DismodFile(None)
```
adolgert: Very happy all these empty braces are disappearing.

src/cascade/dismod/db/metadata.py
```diff
@@ -532,7 +534,7 @@ class Var(Base):
     covariate_id = Column(Integer(), nullable=False)
 
 
-_TYPE_MAP = {str: String, int: Integer, float: Float}
+_TYPE_MAP = {str: String, int: Integer, float: Float, np.dtype("int64"): Integer, np.dtype("float64"): Float}
```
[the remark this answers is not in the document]
adolgert: This changes the type map but doesn't change where it's used. Ah, you want to be able to specify columns by numpy type. I got it. No problem.

src/cascade/dismod/db/wrapper.py
```diff
@@ -187,15 +181,40 @@ def __dir__(self):
         attributes.extend(super().__dir__())
         return attributes
 
+    def update_table_columns(self, table_definition, table):
+        new_columns = table.columns.difference(table_definition.c.keys())
+        new_columns = {c: table.dtypes[c] for c in new_columns}
```
adolgert: The values change here, so it would be easier to notice that if the variable name is different from `new_columns`.

src/cascade/dismod/db/wrapper.py
```diff
@@ -187,15 +181,40 @@ def __dir__(self):
...
+
+        bad_column_names = [c for c in new_columns.keys() if not c.startswith("x_")]
+        if bad_column_names:
+            raise ValueError("Covariate column names must start with 'x_'")
```
adolgert: Add the list of bad column names to the error message.

src/cascade/dismod/db/metadata.py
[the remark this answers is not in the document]
alec-deason: Yep. If we want to keep this tight, that's fine but those are the types that things come in as.

src/cascade/dismod/db/wrapper.py
alec-deason: Good catch, I totally forgot about `empty_table`. This should go through that path.

tests/dismod/db/test_wrapper.py
alec-deason: Hmm. Good point. I don't quite cover this case in my other tests. I'll bring it back.

src/cascade/dismod/db/wrapper.py
alec-deason: Well, here's a readability question. I'm doing this for the sake of not having an absurdly complex one-liner (and instead having two moderately complex one-liners). The first value of `new_columns` is totally ephemeral. Is it more clear to throw that ephemeral value away but maybe muddy peoples' understanding of what the second line does? Or is it more clear to not have the dangling value, which I'll never actually use, and make it clear that there is a transformation in the second line?

src/cascade/dismod/db/wrapper.py
alec-deason: Sure

src/cascade/dismod/db/metadata.py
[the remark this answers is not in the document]
adolgert: I mean that np.dtype("int64") is the same as np.int64, right?

src/cascade/dismod/db/wrapper.py
adolgert: I was confused about why this wrote twice to a variable since the first was thrown away. And Code Complete says reuse of names is a source of errors. Thus sayeth Code Complete.

src/cascade/dismod/db/metadata.py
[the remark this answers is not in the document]
alec-deason: Oh, yes.

src/cascade/dismod/db/wrapper.py
alec-deason: Though I continue to object to the appeal to dubious authority, if you were confused then that's meaningful. I've reorganized it a bit so it should be more clear.

PR #38, opened 2018-09-07
feature/DISMODAT-246-bash-scripts-for-install

nrhawkins:
Updated the install.sh to install from the github repo.

Note: install.sh captures the specific singularity image in DISMOD_AT_PATH
For example, we echo this:
"The physical path of the DISMOD_AT executable is /ihme/singularity-images/dismod/dismod_at:0.0.6.img."

Question: where do we need to store this value?  what other files are going to need it?  are they going to be aware of where we store it?  is this a separate task? 

Note: during install, this message was printed in red:
save-results 0.17.2 has requirement tables==3.4.0, but you'll have tables 3.4.4 which is incompatible.
 
Question: do we have plans to specify specific versions?

### DIFF
--- a/scripts/automation/base_python.sh
+++ b/scripts/automation/base_python.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -x
+
+# This script installs miniconda and numpy into the dismod_at space, 
+# and upgrades pip, virtualenv, and pipenv
+
+# This should be 2 or 3.
+PYTHONVERSION=3
+
+CODE=/ihme/code/dismod_at
+WORKDIR="${CODE}/build"
+SCRIPT="Miniconda${PYTHONVERSION}-latest-Linux-x86_64.sh"
+INSTALLDIR="${CODE}/pyenv/versions"
+
+# Any existing python path can interfere with installation, as can
+# other Pythons on the path.
+unset PYTHONPATH
+export PATH=/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin
+
+cd "${WORKDIR}"
+rm -f "${SCRIPT}"
+wget "https://repo.continuum.io/miniconda/${SCRIPT}"
+
+# The version is stored in this way in the file, so find that.
+# NAME:   Miniconda3
+# VER:    4.5.1
+CONDANAME=`awk '/NAME:/ {print tolower($3)}' ${SCRIPT}`
+CONDAVER=`awk '/VER:/ {print $3}' ${SCRIPT}`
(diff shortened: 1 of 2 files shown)


scripts/automation/base_python.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+set -x
+
+# This script installs miniconda and numpy into the dismod_at space, 
+# and upgrades pip, virtualenv, and pipenv
+
+# This should be 2 or 3.
+PYTHONVERSION=3
```
adolgert: This is now always 3, so let's just get the 3 version.

scripts/automation/base_python.sh
```diff
@@ -0,0 +1,44 @@
...
+
+if [ -e  "${CONDAHOME}" ]; then
+  echo Directory "${CONDAHOME}" exists so cannot install.
+  exit 37
+fi
+
+# Run the installer in batch mode.
+bash "${SCRIPT}" -b -p "${CONDAHOME}"
+
+"${CONDAHOME}/bin/conda" install -y numpy[mkl]
```
adolgert: I think we need a few more conda libraries in order to support get_draws and friends. Please add blosc, redis, hdf5, and mysql-connector-c.

scripts/automation/install.sh
```diff
@@ -0,0 +1,120 @@
...
+echo $ENV
+
+# Create the virtual environment for the project code and required packages
+
+"${VIRTUALENV}" "${ENV}"
+
+source "${ENV}/bin/activate"
+pip install --upgrade pip
+cd "${CASCADE_DEVELOP_DIR}"
+pip install .[ihme_databases,testing]
```
adolgert: May as well include the documentation option b/c it makes available to a user an option to build the docs.

scripts/automation/install.sh
```diff
@@ -0,0 +1,120 @@
...
+# for the dismod_at singularity image
+# /share/singularity-images/dismod/current.img
+# /share/singularity-images/dismod_at/current.img
+
+DISMOD_AT_PATH=`readlink -f /share/singularity-images/dismod/current.img`
+if ! [ -f "${DISMOD_AT_PATH}" ] ; then
+    echo "ERROR: the DISMOD_AT_PATH is invalid!"
+    exit 1
+fi
+echo "The physical path of the DISMOD_AT executable is ${DISMOD_AT_PATH}."
```
adolgert: This is the part where we didn't decide how to tell it where dismod_at is located. Hmmm.

adolgert:
Merged on branch where I can write.

PR #41, opened 2018-09-10
Feature/form validation

alec-deason:
This gives us a toolbox for parsing and validating the parameter blobs that EpiViz produces. It also has a sketch of a full validator for the parameters as a whole (with a bunch a parts just blocked out) and a more or less complete validator for the prior distribution sections of the parameters.

There is quite a bit of metaprogramming and magic in this code so I recommend you start from the surface API and work inward to develop a sense of what's going on.

Good starting places are the definition of a SmoothingPrior (which corresponds to one region within a smoothing grid on our side and one row within the 'details' sections of the various smoothing grid widgets on the GUI side): https://github.com/ihmeuw/cascade/pull/41/files#diff-9bec0033c3dcbc83922adec92d9db7ddR15

Also the tests for concrete field types which shows how various features of the system work: https://github.com/ihmeuw/cascade/pull/41/files#diff-596b28133c38e99676c5977743a1b24f

There  will be a follow on PR which uses these tools to actually pull parameters from EpiViz and use them to setup parts of the internal model. But I figure this bit is complex enough that it should go out on its own.


src/cascade/model/priors.py
```diff
@@ -44,17 +48,17 @@ def __repr__(self):
...
 def _validate_standard_deviation(standard_deviation):
     if standard_deviation < 0:
-        raise ValueError("Standard deviation must be positive")
+        raise PriorError("Standard deviation must be positive")
 
 
 def _validate_nu(nu):
     if nu < 0:
-        raise ValueError("Nu must be positive")
+        raise PriorError("Nu must be positive")
```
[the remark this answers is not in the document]
alec-deason: Yes, that would make this clearer.

src/cascade/input_data/configuration/form.py
```diff
@@ -0,0 +1,134 @@
...
+    max_num_iter = Dummy()
+    print_level = Dummy()
+    accept_after_max_steps = Dummy()
+    tolerance = Dummy()
+    students_dof = Dummy()
+    log_students_dof = Dummy()
+    eta = Dummy()
+    data_eta_by_integrand = Dummy()
+    data_density_by_integrand = Dummy()
+    config_version = Dummy()
```
adolgert: This one's a dummy? We might want to read the config version early, which raises the question of whether we can version this set of parameters in any real way. Yes, we could. We'd have a root for each version. And we'd have to write code to promote settings from one root to fill out the settings in the next root. Seems unlikely we get to include that as a feature, but that's how it would work.

src/cascade/input_data/db/__init__.py
```diff
@@ -1,3 +1,13 @@
+try:
+    import db_queries
+except ImportError:
+
+    class DummyDBQueries:
+        def __getattr__(self, name):
+            raise ImportError(f"Required package db_queries not found")
+
+    db_queries = DummyDBQueries()
```
adolgert: Huh. Interesting. This isn't a fake, a mock, or a stub. It's more of an explode-on-impact.

src/cascade/input_data/db/__init__.py
```diff
@@ -1,3 +1,13 @@
...
+
 AGE_GROUP_SET_ID = 12
 
 GBD_ROUND_ID = 5
```
adolgert: I know this isn't the code under question, but these need to be somewhere else. I'd prefer to have them all the way up in the admin interface on EpiViz.

src/cascade/input_data/db/__init__.py
adolgert: Hey, could we do this for `db_tools` and `save_results` in the same place?

src/cascade/input_data/db/bundle.py
```diff
@@ -257,7 +257,7 @@ def _normalize_measures(data):
     """Transform measure_ids into canonical measure names
     """
     data = data.copy()
-    data["measure"] = data.measure_id.apply(MEASURES.get)
+    data["measure"] = data.measure_id.apply(lambda k: MEASURES[k])
```
adolgert: Why go for the lambda? It's a stylistic choice of some sort.

tests/core/form/test_abstract_form.py
```diff
@@ -0,0 +1,181 @@
...
+    assert f.my_field == 10
+    assert f.my_inner_form.my_inner_field == 100
+
+    f = MyForm({"my_field": "10", "my_inner_form": {"my_inner_field": "eaoeao"}})
+    errors = f.validate_and_normalize()
+    assert set(errors) == {("my_inner_form.my_inner_field", "Invalid integer value 'eaoeao'")}
+
+    f = MyForm({"my_field": None, "my_inner_form": {}})
+    errors = f.validate_and_normalize()
+    assert set(errors) == {
```
adolgert: I like that it returns multiple errors. I know that's what we need. It's just nice.

tests/core/form/test_fields.py
```diff
@@ -0,0 +1,163 @@
...
+def test_StringListField__successful_normalization(form_with_string_list):
+    f = form_with_string_list({"ints_field": "1 2 3 4 5"})
+    f.validate_and_normalize()
+    assert f.ints_field == [1, 2, 3, 4, 5]
+
+
+def test_StringListField__failed_validation(form_with_string_list):
+    f = form_with_string_list({"ints_field": "1 2 three 4 five"})
+    assert f.validate_and_normalize() == [
+        ("ints_field", "Errors in items: [Invalid int value 'three', Invalid int value 'five']")
```
adolgert: What a nice error message.

src/cascade/core/form/abstract_form.py
```diff
@@ -0,0 +1,291 @@
...
+    """
+
+    def __repr__(self):
+        return "NO_VALUE"
+
+
+NO_VALUE = NoValue()
+
+
+class Node:
```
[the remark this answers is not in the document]
alec-deason: Yeah, Node isn't great. It's just a container for behavior shared by both Forms and Fields. I'll think of something tighter.

src/cascade/core/form/abstract_form.py
```diff
@@ -0,0 +1,291 @@
...
+    Validation for forms happens in two stages. First all the form's fields and
+    sub forms are validated. If none of those have errors, then the form is
+    known to be in a consistent state and it's `_full_form_validation` method
+    is run to finalize validation. If any field or sub form is invalid then
+    this form's `_full_form_validation` method will not be run because the form
+    may be in an inconsistent state.
+
+    Simple forms will be valid if all their fields are valid but more complex
+    forms will require additional checks across multiple fields which are
+    handled by `_full_form_validation`.
```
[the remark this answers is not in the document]
alec-deason: I have a convention in my head where hooks that are intended to be overridden by subclasses but not used by things outside the hierarchy get flagged with the underscore. Is that not a thing?

src/cascade/input_data/configuration/form.py
alec-deason: Things I don't know what to do with are Dummy. That's true of this one. How do I use it? Aside from a scheme like you describe which I agree we aren't doing. Also, at the moment, I have no idea what EpiViz means by config version.

src/cascade/input_data/db/__init__.py
adolgert: Yeah. We need a parameters section. And I want to work to make it small. Even if they are hidden parameters in the EpiViz interface, that would make sense.

src/cascade/input_data/db/bundle.py
adolgert: Ah. That works. Could also use ``MEASURES.__getitem__``. Python is against lambdas. I need to either get on board with that or get on board against it.

PR #47, opened 2018-09-17
Feature/covariates representation

adolgert:
I think I need to work on the testing but would love a review, if possible.


src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+    """
+    Establishes a reference value for a covariate column on input data
+    and in output data. It is possible to create a covariate column with
+    nothing but a name, but it must have a reference value before it
+    can be used in a model.
+
+    Args:
+        column_name (str): Name of hte column in the input data.
+        reference (float): Optional reference where covariate has no effect.
+        max_difference (float, optional):
```
alec-deason: Inconsistent way of marking optional-ness.

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+    def max_difference(self):
+        return self._max_difference
+
+    @max_difference.setter
+    def max_difference(self, difference):
+        if difference is None:
+            self._max_difference = None
+        else:
+            diff = float(difference)
+            if diff <= 0:
```
alec-deason: Brad says "This value must be greater than or equal zero"

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+    def reference(self, ref):
+        self._reference = float(ref)
+
+    @property
+    def max_difference(self):
+        return self._max_difference
+
+    @max_difference.setter
+    def max_difference(self, difference):
+        if difference is None:
```
alec-deason: Null == +infinity is Brad's convention but isn't necessarily obvious. Do we want to carry that into our model or would it be clearer to make people specify infinity? Honest question. I don't know which is clearer.

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+    def __eq__(self, other):
+        return (self._name == other.name and self._reference == other._reference
+                and self._max_difference == other._max_difference)
+
+
+class CovariateMultiplier:
+    """
+    A covariate multiplier makes a given covariate column the predictor
+    of a model variable, where the model variable can be one of a
+    rate, an integrand's value, or an integrand's standard deviation.
```
alec-deason: The rate's value?

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+
+    This class is only two-thirds of the definition.
+    The covariate column has to be attached to a particular rate
+    or measured value or measured standard deviation.
+    """
+    def __init__(self, covariate_column, smooth):
+        """
+        Args:
```
alec-deason: Shouldn't this block be in the class docstring?

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def __repr__(self):
+        return f"CovariateColumn({self.name}, {self.reference}, {self.max_difference})"
+
+    def __eq__(self, other):
+        return (self._name == other.name and self._reference == other._reference
+                and self._max_difference == other._max_difference)
+
+
+class CovariateMultiplier:
```
alec-deason: Why express this as a linking object rather than just a list of covariates on the relevant smooth?

src/cascade/dismod/serialize.py
```diff
@@ -391,25 +405,58 @@ def _smooth_row(name, smooth, grid, prior_id_func):
...
+        for val_mul in val_integrand.value_covariate_multipliers:
+            yield val_mul, "meas_value", val_integrand
+
+    # γ
+    for std_integrand in context.outputs.integrands:
+        for std_mul in std_integrand.std_covariate_multipliers:
+            yield std_mul, "meas_std", std_integrand
+
+
+def smooth_iter(context):
```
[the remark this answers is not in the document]
adolgert: I wonder about that for all of the collect_* functions. This is a version of a collect_* function.

src/cascade/model/covariates.py
```diff
@@ -0,0 +1,101 @@
...
+"""
+import logging
+from cascade.model.rates import Smooth
+
+
+CODELOG = logging.getLogger(__name__)
+MATHLOG = logging.getLogger(__name__)
+
+
+class CovariateColumn:
```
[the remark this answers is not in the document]
adolgert: Yeah. These could be the primary thing.

src/cascade/dismod/serialize.py
```diff
@@ -436,17 +483,68 @@ def smooth_id_func(smooth):
...
+        "covariate_name": [col.name for col in cols],
+        "reference": np.array([col.reference for col in cols], dtype=np.float),
+        "max_difference": np.array([col.max_difference for col in cols], dtype=np.float),
+    })
+    if covariate_columns["reference"].isnull().any():
+        null_references = list()
+        for check_ref_col in cols:
+            if not isinstance(check_ref_col.reference, Real):
+                null_references.append(check_ref_col.name)
+        raise RuntimeError(f"Covariate columns without reference values {null_references}")
```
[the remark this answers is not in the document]
adolgert: I changed my mind after talking with Gregg and didn't update docs.

src/cascade/model/covariates.py
adolgert: ? What the heck does it mean if it's zero? OK, Brad.

src/cascade/model/covariates.py
adolgert: I don't think his +Inf makes any sense to bring to the modeler. None.

src/cascade/model/covariates.py
adolgert: THIS is the big question and why I posted the PR a little early. The smooth isn't the intrinsic statistical object. It's a "support," so it's not primary. But you could attach to the covariate the things it multiplies instead of attaching to the thing it multiplies the covariate.

src/cascade/dismod/serialize.py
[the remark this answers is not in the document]
alec-deason: Valid. This whole system could do with a shuffle. I'm very unclear on where the proper boundaries are.

src/cascade/model/covariates.py
alec-deason: Cool

src/cascade/model/covariates.py
alec-deason: That makes sense to me. I'm just nervous about introducing weird mapping table type concepts where we don't need them. Hanging a list of smooths off of the covariate objects seems reasonable to me.

src/cascade/dismod/serialize.py
```diff
@@ -436,17 +483,68 @@ def smooth_id_func(smooth):
...
+
+        cm_rows.append(dict(
+            mulcov_id=cidx,
+            mulcov_type=kind,
+            rate_id=rate_id,
+            integrand_id=integrand_id,
+            covariate_id=cov_col_id_func(cov_mul.column),
+            smooth_id=smooth_id_func(cov_mul.smooth),
+        ))
+    mul_cov = pd.DataFrame.from_records(
```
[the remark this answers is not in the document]
adolgert: It needed the columns command. I added that and the ``from_records.`` OK, gotten rid-of.

src/cascade/dismod/serialize.py
```diff
@@ -391,25 +405,58 @@ def _smooth_row(name, smooth, grid, prior_id_func):
     }
 
 
+def covariate_multiplier_iter(context):
```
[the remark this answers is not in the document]
adolgert: I don't know. We should figure that out. Separately, I think, from the PR.

src/cascade/dismod/serialize.py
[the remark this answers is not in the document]
adolgert: I fixed the model code.

src/cascade/dismod/serialize.py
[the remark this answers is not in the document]
alec-deason: Sure

PR #48, opened 2018-09-18
Feature/acceptance test

adolgert:
The goal is to define tests according to when we run them:

1. Default is not to use IHME resources or do long tests.
2. The `--ihme-db` flag turns on availability of IHME databases. This is for laptop tests within VPN.
3. To run acceptance tests, go to the acceptance directory and run pytest.

What's complicated: In order to use IHME databases, you now have to include a fixture called "ihme_db". If you don't, then the tests will fail.

### DIFF
--- a/src/cascade/input_data/db/__init__.py
+++ b/src/cascade/input_data/db/__init__.py
@@ -1,12 +1,4 @@
-try:
-    import db_queries
-except ImportError:
-
-    class DummyDBQueries:
-        def __getattr__(self, name):
-            raise ImportError(f"Required package db_queries not found")
-
-    db_queries = DummyDBQueries()
+from cascade.input_data.db import module_proxy
 
 AGE_GROUP_SET_ID = 12
 
@@ -15,3 +7,11 @@ GBD_ROUND_ID = 5
 METRIC_IDS = {"per_capita_rate": 3}
 
 MEASURE_IDS = {"deaths": 1}
+
+db_queries = module_proxy.ModuleProxy("db_queries")
+db_tools = module_proxy.ModuleProxy("db_tools")
+save_results = module_proxy.ModuleProxy("save_results")
+
+
+def disable_databases():
+    module_proxy.BLOCK_SHARED_FUNCTION_ACCESS = True
--- a/src/cascade/input_data/db/__init__.py
+++ b/src/cascade/input_data/db/__init__.py
@@ -11,7 +11,3 @@ MEASURE_IDS = {"deaths": 1}
 db_queries = module_proxy.ModuleProxy("db_queries")
 db_tools = module_proxy.ModuleProxy("db_tools")
 save_results = module_proxy.ModuleProxy("save_results")
-
-
-def disable_databases():
-    module_proxy.BLOCK_SHARED_FUNCTION_ACCESS = True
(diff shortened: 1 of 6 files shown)


src/cascade/input_data/db/__init__.py
```diff
@@ -15,3 +7,7 @@ def __getattr__(self, name):
 METRIC_IDS = {"per_capita_rate": 3}
 
 MEASURE_IDS = {"deaths": 1}
+
+db_queries = module_proxy.ModuleProxy("db_queries")
```
alec-deason: I would put module_proxy in cascade.core.db and also have these wrapped modules there as well. That way other modules can go `from cascade.core.db import db_queries` without needing to know the gory details.

src/cascade/input_data/db/module_proxy.py
```diff
@@ -0,0 +1,39 @@
...
+        except ModuleNotFoundError:
+            self._module = None
+
+    def __getattr__(self, name):
+        if BLOCK_SHARED_FUNCTION_ACCESS:
+            raise ModuleNotFoundError(
+                f"Illegal access to module {self.name}. Are you trying to use "
+                f"the shared functions in a unit test?")
+
+        return getattr(self._module, name)
```
alec-deason: It would be nice if this gave a clear error message in the case where self._module is None

tests/conftest.py
```diff
@@ -18,3 +21,25 @@ def mock_ezfuncs(mocker):
 @pytest.fixture
 def mock_database_access(mock_ezfuncs):
     return {"cursor": mock_ezfuncs.get_connection().cursor(), "connection": mock_ezfuncs.get_connection()}
+
+
+def pytest_addoption(parser):
```
alec-deason: Neat, but dear lord why is this a magic function name and not a decorator off the pytest module or something like that that's consistent with the rest of their API. Not your code, I know. Bleck.

src/cascade/input_data/db/module_proxy.py
```diff
@@ -0,0 +1,39 @@
...
+
+        self.name = module_name
+        try:
+            self._module = importlib.import_module(module_name)
+        except ModuleNotFoundError:
+            self._module = None
+
+    def __getattr__(self, name):
+        if BLOCK_SHARED_FUNCTION_ACCESS:
+            raise ModuleNotFoundError(
```
alec-deason: I'm not sure what the right exception here is but this one feels wrong. The module isn't necessarily unfindable, you just aren't allowed to use it.

tests/conftest.py
```diff
@@ -18,3 +21,25 @@ def mock_ezfuncs(mocker):
...
+    return IhmeDbFuncArg(request)
+
+
+class IhmeDbFuncArg:
+    """
+    Uses a pattern from https://pytest.readthedocs.io/en/2.0.3/example/attic.html
+    """
+    def __init__(self, request):
+        if not request.config.getoption("ihme_db"):
+            pytest.skip(f"specify --ihme to run tests requiring Central Comp databases")
```
nrhawkins: Should this be: "specify --ihme-db ?

adolgert: We've seen this once before in parsing with argparse. The flag is always using middle-dashes, but then it gets converted into an underscore when it's a Python variable because dashes aren't allowed.

alec-deason: I think she's talking about the skip message where it calls the flag "--ihme"

PR #54, opened 2018-10-04
Build integrands from gbd demographics

alec-deason:


src/cascade/core/context.py
```diff
@@ -36,51 +37,17 @@ def __iter__(self):
...
-    __slots__ = ["integrands"]
-
-    def __init__(self):
-        self.integrands = _Integrands()
-
-
-class _Integrand:
-    __slots__ = ["name", "age_ranges", "time_ranges", "value_covariate_multipliers", "std_covariate_multipliers"]
+class _IntegrandCovariateMultiplier:
+    __slots__ = ["name", "value_covariate_multipliers", "std_covariate_multipliers"]
```
adolgert: Darn it. There are 3 types of covariate multipliers, not two. You can have one on the value of an integrand, the stdev of an integrand OR on the value of a rate. So if you put a covariate multiplier on "remission" value, that's different from a covariate multiplier on "iota" value. I don't think this setup expresses the rate covariate multipliers. The avgint table can produce not only integrands, which is what the name implies, but rates, too.

src/cascade/core/context.py
```diff
@@ -36,51 +37,17 @@ def __iter__(self):
...
-        for name in self.__slots__:
-            setattr(self, name, _Integrand(name))
-
-    def __iter__(self):
-        for name in self.__slots__:
-            yield getattr(self, name)
-
-
 class ModelContext:
+    __slots__ = ["parameters", "input_data", "rates", "average_integrand_cases", "integrand_covariate_multipliers"]
```
adolgert: I like your carefulness with slots. It avoids attaching things randomly.

src/cascade/dismod/db/metadata.py
```diff
@@ -534,7 +534,14 @@ class Var(Base):
     covariate_id = Column(Integer(), nullable=False)
 
 
-_TYPE_MAP = {str: String, int: Integer, float: Float, np.dtype("int64"): Integer, np.dtype("float64"): Float}
+_TYPE_MAP = {
```
adolgert: I think this map is my fault. I'm realizing I need to make the name more specific. It's a map from pandas types to sqlalchemy types, I think.

src/cascade/dismod/serialize.py
```diff
@@ -442,15 +423,13 @@ def covariate_multiplier_iter(context):
...
-    for val_integrand in context.outputs.integrands:
-        for val_mul in val_integrand.value_covariate_multipliers:
-            yield val_mul, "meas_value", val_integrand
-
-    # γ
-    for std_integrand in context.outputs.integrands:
-        for std_mul in std_integrand.std_covariate_multipliers:
-            yield std_mul, "meas_std", std_integrand
+    for integrand in context.integrand_covariate_multipliers.values():
+        # β
```
adolgert: I put this unicode character in here. Think we're safe with those? Recommendation is not to use them b/c they still sometimes get misinterpreted.

src/cascade/executor/no_covariate_main.py
```diff
@@ -207,9 +207,21 @@ def internal_model(model, inputs):
     for rate in config.non_zero_rates:
         getattr(model.rates, rate).parent_smooth = default_smoothing
 
+    avgint_rows = []
```
adolgert: We have both mains() to deal with, the no_covariate and the epiviz. I forgot that. Nice catch.

src/cascade/input_data/configuration/builder.py
```diff
@@ -23,13 +22,16 @@
...
 
 
-def squared(x): return np.power(x, 2)
+def squared(x):
+    return np.power(x, 2)
 
 
-def scale1000(x): return x * 1000
+def scale1000(x):
+    return x * 1000
```
adolgert: Didn't like the single line? ;)

src/cascade/input_data/configuration/builder.py
```diff
@@ -74,41 +76,6 @@ def unique_country_covariate_transform(configuration):
         yield cov_id, list(sorted(cov_transformations))
 
 
-def exclude_integrands_without_extents(integrands):
```
adolgert: Great. This code was wrong.

src/cascade/model/integrands.py
```diff
@@ -0,0 +1,42 @@
...
+            "time_upper": time_upper,
+            # Assuming using the first set of weights, which is constant.
+            "weight_id": 0,
+            "node_id": execution_context.parameters.location_id,
+            "x_sex": sex,
+        }
+        for integrand in IntegrandEnum
+        for age_lower, age_upper in age_ranges
+        for time_lower, time_upper in time_ranges
+        for sex in [-0.5, 0.5]
```
adolgert: This is damned beautiful.

src/cascade/core/context.py
alec-deason: I do it because I keep getting it wrong and on these objects that exist only to hold attributes, that sucks.

src/cascade/dismod/serialize.py
alec-deason: As comments or as variable names? I did once write a chunk of code that used fancy unicode variable names. It was kind of neat because I could almost copy and paste from math articles and expect it to run but they're too hard to type. No idea about misinterpretation.

src/cascade/model/integrands.py
alec-deason: Has Falko told you about his "stars" system. 0 stars means he thinks the code is fine, 3 stars means he's sure I will yell when he shows it to me. I was giving this deeply nested comprehension 3 stars because it looks like Haskell. But I'm glad you like it. I'm pretty sure code complete has a chapter about how you shouldn't do this.

alec-deason:
Thank you for going through all these.

PR #55, opened 2018-10-04
Bugfix or something/make diabetes go

alec-deason:
This is a dump of the stuff I've been working on. Mostly just a bunch of bug fixes.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -4,12 +4,23 @@ from pathlib import Path
 from pprint import pformat
 import json
 
+import pandas as pd
+import numpy as np
+
 from cascade.executor.argument_parser import DMArgumentParser
-from cascade.input_data.db.demographics import get_age_groups, get_years
+from cascade.input_data.db.demographics import age_groups_to_ranges
 from cascade.dismod.db.wrapper import _get_engine
+from cascade.dismod.db.metadata import DensityEnum
 from cascade.testing_utilities import make_execution_context
 from cascade.input_data.db.configuration import settings_for_model
-from cascade.executor.no_covariate_main import bundle_to_observations
+from cascade.input_data.db.csmr import load_csmr_to_t3
+from cascade.input_data.db.asdr import load_asdr_to_t3
+from cascade.input_data.db.mortality import (
+    get_cause_specific_mortality_data,
+    get_age_standardized_death_rate_data,
+)
+from cascade.model.integrands import integrand_grids_from_gbd
+from cascade.executor.no_covariate_main import bundle_to_observations, build_constraint
 from cascade.executor.dismod_runner import run_and_watch, DismodATException
 from cascade.input_data.configuration.form import Configuration
 from cascade.input_data.db.bundle import bundle_with_study_covariates, freeze_bundle
@@ -23,6 +34,7 @@ from cascade.input_data.configuration.builder import (
 )
 
 CODELOG = logging.getLogger(__name__)
+MATHLOG = logging.getLogger("cascade_a.math.runner")
 
 
 def load_settings(meid=None, mvid=None, settings_file=None):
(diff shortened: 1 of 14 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -49,37 +61,73 @@ def execution_context_from_settings(settings):
         model_version_id=settings.model.model_version_id,
         model_title=settings.model.title,
         gbd_round_id=settings.gbd_round_id,
+        bundle_id=settings.model.bundle_id,
+        add_csmr_cause=settings.model.add_csmr_cause,
```
adolgert: add csmr cause seems like configuration? Let's lay these out and pick what goes where and _why._

src/cascade/executor/epiviz_runner.py
```diff
@@ -49,37 +61,73 @@ def execution_context_from_settings(settings):
...
+        location_id=settings.model.drill_location,
+    )
+
+
+def meas_bounds_to_stdev(df):
+    df["standard_error"] = (df.meas_upper - df.meas_lower) / (2 * 1.96)
+    df["standard_error"] = df.standard_error.replace({0: 1e-9})
```
adolgert: These are the kind of thing to put in the math log.

adolgert: And to set in parameters for epiviz_runner.

src/cascade/executor/epiviz_runner.py
```diff
@@ -49,37 +61,73 @@ def execution_context_from_settings(settings):
...
+    df = df.rename(columns={"meas_value": "mean"})
+    df["density"] = DensityEnum.gaussian
+    df["weight"] = "constant"
+    return df.drop(["meas_lower", "meas_upper"], axis=1)
+
+
+def add_mortality_data(model_context, execution_context):
+    csmr = meas_bounds_to_stdev(
```
adolgert: Do we add mortality as an observation, or is the "add mortality data" drop-down really about setting mtother from that particular mortality? Maybe that drop-down means to do both, add the measure and add mtother. I need to look at DismodODE.

src/cascade/executor/epiviz_runner.py
```diff
@@ -49,37 +61,73 @@ def execution_context_from_settings(settings):
...
+        age_groups_to_ranges(execution_context, get_cause_specific_mortality_data(execution_context))
     )
+    csmr["measure"] = "mtspecific"
+    model_context.input_data.observations = pd.concat([model_context.input_data.observations, csmr])
+
+
+def add_omega_constraint(model_context, execution_context):
+    asdr = meas_bounds_to_stdev(
```
adolgert: Oh, here it is, the adding of mtother. I should read and then comment.

src/cascade/executor/epiviz_runner.py
```diff
@@ -49,37 +61,73 @@ def execution_context_from_settings(settings):
...
+    freeze_bundle(execution_context, execution_context.parameters.bundle_id)
+    load_csmr_to_t3(execution_context)
+    load_asdr_to_t3(execution_context)
 
     bundle, study_covariates = bundle_with_study_covariates(
         execution_context, bundle_id=model_context.parameters.bundle_id
     )
+    bundle = bundle.query("location_id == @execution_context.parameters.location_id")
     model_context.input_data.observations = bundle_to_observations(model_context.parameters, bundle)
+    mask = model_context.input_data.observations.standard_error > 0
```
adolgert: maybe name it mask_zero_stderr?

src/cascade/executor/epiviz_runner.py
```diff
@@ -136,16 +178,30 @@ def main():
...
+        MATHLOG.error(f"Form data: {pformat(e.form_data)}")
+        MATHLOG.error(f"Form validation errors: {pformat(e.form_errors)}")
         exit(1)
     except Exception:
-        CODELOG.exception(f"Uncaught exception in {os.path.basename(__file__)}")
-        raise
+        if args.pdb:
+            import pdb
+            import traceback
+
```
adolgert: In the right direction here.

src/cascade/input_data/configuration/builder.py
```diff
@@ -256,16 +262,24 @@ def covariate_to_measurements_nearest_favoring_same_year(measurements, covariate
...
-        covariates[["age_lower", "age_upper"]].mean(axis=1) / 240,
-        covariates[["time_lower", "time_upper"]].mean(axis=1),
-        covariates["x_sex"]
-    )))
-    _, indices = tree.query(list(zip(
-        measurements[["age_lower", "age_upper"]].mean(axis=1) / 240,
-        measurements[["time_lower", "time_upper"]].mean(axis=1),
-        measurements["x_sex"],
-    )))
+    tree = spatial.KDTree(
```
adolgert: This area might conflict with Natalie's work. Looks like a reformatting, not a change in what it does.

src/cascade/input_data/db/bundle.py
```diff
@@ -13,7 +13,16 @@
 CODELOG = logging.getLogger(__name__)
 
 # FIXME: There is a shared function that get's the official mapping, I think. Or an sql query at least.
-MEASURES = {6: "incidence", 9: "mtexcess", 5: "prevalence"}
+MEASURES = {
```
adolgert: The map to dismod names is currently inside the db.metadata as an Enum. I hope not to need it far from there.

tests/model/test_integrands.py
```diff
@@ -0,0 +1,25 @@
+import pandas as pd
+
+from cascade.testing_utilities import make_execution_context
+from cascade.core.context import ModelContext
+
+from cascade.model.integrands import integrand_grids_from_gbd
+
+
+def test_integrand_grids_from_gbd(mocker):
```
adolgert: a test! Yes, we have cleanup to do.

src/cascade/executor/epiviz_runner.py
alec-deason: Yes, please.

src/cascade/executor/epiviz_runner.py
alec-deason: I believe there are rules that ODE and to some extent Greg use when deciding what mortality data to add and/or synthesize under what circumstances. That's a thing I don't fully understand yet.

src/cascade/executor/epiviz_runner.py
alec-deason: This is a bit of tooling I love. I figured it out for vivarium and stole it from there for us. It should probably go in your argparse code but that didn't have a clear place to put it and I didn't have time to explore.

tests/model/test_integrands.py
alec-deason: Yeah, they are thin on the ground for this stuff. Both because I was struggling to get through the weeds and because this is the kind of code that is anti-fun to test. But we should fix that.

PR #57, opened 2018-10-16
Bugfix/anxiety submit

adolgert:
This makes meid=1989, anxiety work. No covariates. Needs to edit the EpiViz model version a bit or else it will fail. What's in here:
- When you load dmcascade from a settings file, it now goes to get the mvid.
- New "gbd" package changed its default gbd_round_id to 6, so started passing it into the covariates function.
- Bundle translation was missing some of the measures. Decided to put incidence into the list and code it as Tincidence, but throw a warning.


src/cascade/input_data/db/configuration.py
```diff
@@ -2,30 +2,111 @@
...
+    """
+    Given a meid, get settings for the latest corresponding mvid.
+
+    Args:
+        modelable_entity_id (int,str): The MEID.
+    Returns:
+        dict: Settings as a JSON dictionary of dictionaries.
+        int: Model version ID that is the latest one associated with this meid.
+    """
+    ec = make_execution_context(modelable_entity_id=modelable_entity_id)
```
alec-deason: This is dangerous because it looses the prod/dev distinction that may be in place in the outer code. The context should be passed in. This is unlikely to cause problems now but it's brittle and bug prone.

alec-deason: I know that you are just copying the pattern I established, but I was lazy and wrong.

src/cascade/input_data/db/configuration.py
```diff
@@ -2,30 +2,111 @@
...
+    Args:
+        settings_file (str): Model version ID.
+    Returns:
+        dict: Settings as a JSON dictionary of dictionaries.
+        int: Model version ID, either found in file or latest from meid.
+    """
+    with open(str(settings_file), "r") as f:
+        raw_settings = json.load(f)
+    if "model" in raw_settings and "modelable_entity_id" in raw_settings["model"]:
+        ec = make_execution_context(modelable_entity_id=raw_settings["model"]["modelable_entity_id"])
```
alec-deason: Again, this should take an ExecutionContext as an argument and then add the mvid to it.

src/cascade/input_data/db/configuration.py
```diff
@@ -2,30 +2,111 @@
...
+    if "model_version_id" in raw_settings["model"]:
+        mvid = raw_settings["model"]["model_version_id"]
+        MATHLOG.info(f"Using mvid {mvid} from the settings file.")
+    else:
+        mvid = latest_model_version(ec)
+        MATHLOG.info(f"Using mvid {mvid} from latest model version.")
+    return raw_settings, mvid
+
+
+def json_settings_to_frozen_settings(raw_settings, mvid=None):
```
alec-deason: Yes, good to extract this from the executor code.

src/cascade/input_data/db/configuration.py
adolgert: You need the execution context to get the settings, but you need the settings to make the whole execution context. It looks like it would help to have one execution context that is the environment of the application. That gets passed into here. Then have a different one, or a different part of it, that is says where we are in the cascade hierarchy. That's what gets made. I don't want to go after that in this PR. That's a reorganization.

src/cascade/input_data/db/configuration.py
adolgert: I thought the executor was the thing that ran Dismod-AT. Now it's any kind of main() plus the thing that ran Dismod-AT. I want to reorganize.

src/cascade/input_data/db/configuration.py
adolgert: NVMD. I made the execution context and then appended to it as I went.

PR #64, opened 2018-10-19
Bugfix/covariates debug model

adolgert:
Well, is this what you were up to, or different? I was working through meid=2005.


src/cascade/input_data/__init__.py
```diff
@@ -1,2 +1,6 @@
 from cascade.core.log import getLoggers
 CODELOG, MATHLOG = getLoggers(__name__)
+
+
+class InputDataError(Exception):
+    """These are errors that result from faults in the input data."""
```
alec-deason: Good!

PR #65, opened 2018-10-19
Feature/improve mathlog messages

adolgert:
This adds some mathlog messages in places I thought they would help.
The big machinery change is that errors collected from the settings validation now show up using the same names that are used in the EpiViz GUI.

### DIFF
--- a/src/cascade/core/form/abstract_form.py
+++ b/src/cascade/core/form/abstract_form.py
@@ -65,11 +65,11 @@ class FormComponent:
 
     _children = None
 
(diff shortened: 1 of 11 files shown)


src/cascade/core/form/abstract_form.py
```diff
@@ -129,18 +135,18 @@ def validate_and_normalize(self, instance, root=None):
 
         Returns:
             [(str, str)]: a list of error messages with path strings
```
alec-deason: 3 strs not 2

PR #67, opened 2018-10-23
Feature/study covariates add to measurements

adolgert:
This hasn't run all the way through because the input covariate data is incorrect. I could use better testing. I'd like to PR it because there shouldn't be any more work on features. The main feedback I'd like is suggestions on how to be clearer about stages of transformation of the data from settings and db to the DismodFile and about suggestions for constructing tests that would work despite the databases being down.


docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+=========================
+
+.. _epiviz-overview-introduction:
+
+Introduction
+------------
+EpiViz-AT is a web page on http://epimodeling-web-d01.ihme.washington.edu/at
+or http://epimodeling-web-p01.ihme.washington.edu/at. After you fill out
+that web page and hit submit, the web server runs a program on the cluster.
+We call that program ``epiviz_runner``. It installed on the IHME cluster,
```
nrhawkins: It installed -> It is installed

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+ 4. Find the list of study covariates associated with this bundle,
+    both by covariate id and by name.
+
+ 5. If data is in demographic notation, then the bundle end times
+    will be wrong, but the code prints a message to say whether it adds
+    a year to the end of an age region or time region.
+
+ 6. It assigns a set of weights to the measurements. These are
+    meant to give relative importance according to changes in
+    population size. These are all 1 if the weight method is "constant."
```
nrhawkins: "constant." -> "constant". i.e.     ."   ->   ".

src/cascade/executor/epiviz_runner.py
```diff
@@ -173,6 +204,21 @@ def has_random_effects(model):
...
+
+     3. Load Settings, which are the values you entered into EpiViz-AT on the web page.
+
+     4. Using the Settings and the Execution Context, fill out the data and
+        any parameters for Dismod-AT, as described in
+        :ref:`build-model-from-epiviz-settings`.
+
+     5. Put that data into a file and run Dismod-AT on that file.
+    """
     ec = make_execution_context()
```
nrhawkins: Just saying, there is a "1. Parse ..." above, and just after the comment, we start with "2. Construct an EC"
I see in the next step, we use args.

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+a table of measurements in the Dismod-AT input file.
+
+ 1. Retrieve the measurement bundle by ``bundle_id`` from the
+    epi database.
+
+ 2. Convert each record in the measurement bundle from its
+    GBD measure ID to a Dismod measure, according to the
+    `bundle id map <https://github.com/ihmeuw/cascade/blob/develop/src/cascade/input_data/configuration/id_map.py>`_.
+    If the incoming data uses incidence, instead of S-incidence
+    or T-incidence, reject it.
```
nrhawkins: Should we comment on "why" we reject it?

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+
+ 3. Convert sex id from 1, 2, 3, 4 to male, female, both, or unspecified.
+
+ 4. Find the list of study covariates associated with this bundle,
+    both by covariate id and by name.
+
+ 5. If data is in demographic notation, then the bundle end times
```
nrhawkins: It could be useful to know more about this "if".   It sounds like the bundle data can contain a mix of demographic notation and not demographic notation?  Or that some bundles use demographic notation and some don't, but no mixing?  Do we have any "why's" for that?  We decide if it is demographic notation or not based on whether or not the start time = the end time ( => demographic notation) ?  The log message is written to say that 'hey, we found demographic notation here!'?

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+
+At the start of this process, the measurement bundle is on
+the Epi database. By the end of this process, there is
+a table of measurements in the Dismod-AT input file.
+
+ 1. Retrieve the measurement bundle by ``bundle_id`` from the
+    epi database.
+
+ 2. Convert each record in the measurement bundle from its
+    GBD measure ID to a Dismod measure, according to the
```
nrhawkins: Should we comment on why Dismod doesn't use the GBD measure ID?

docs/epiviz/overview.rst
nrhawkins: maybe this could be: "5. Adjust age and time end values, if the data is in demographic notation.  ... Log these adjustments so that <someone can see them, because someone will want to know?>"

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+ 5. If data is in demographic notation, then the bundle end times
+    will be wrong, but the code prints a message to say whether it adds
+    a year to the end of an age region or time region.
+
+ 6. It assigns a set of weights to the measurements. These are
+    meant to give relative importance according to changes in
+    population size. These are all 1 if the weight method is "constant."
+
+ 7. All observations with standard error at zero are removed,
+    unless they are relative risks.
```
nrhawkins: I guess there's more after this?  The next steps after "convert bundle"?

src/cascade/executor/epiviz_runner.py
```diff
@@ -131,6 +158,10 @@ def model_context_from_settings(execution_context, settings):
     add_mortality_data(model_context, execution_context)
     add_omega_constraint(model_context, execution_context)
```
nrhawkins: I guess both of these "add's" update "model_context.input_data.observations"
Are "mtspecific" and "mtall" being added to observations? 
Maybe: add_cause_specific_mortality_data_to_observations?
Maybe: add_mtall_and_omega-constraint_to_observations?
Or maybe follow the same approach as for bundle_to_observations?
so that, model_context.input_data.observations = observations_with_mtspecific?

src/cascade/executor/epiviz_runner.py
```diff
@@ -131,6 +158,10 @@ def model_context_from_settings(execution_context, settings):
...
     model_context.average_integrand_cases = make_average_integrand_cases_from_gbd(execution_context)
+    add_special_study_covariates(study_covariate_records, bundle, model_context.average_integrand_cases)
+
+    fixed_effects_from_epiviz(model_context, study_covariate_records, execution_context, settings)
+    random_effects_from_epiviz(model_context, settings)
```
nrhawkins: Seeing these (FE's and RE's) makes me wonder if there is a model_context.fixed_effects and model_context.random_effects?  Maybe add details to the comments above Steps 6 and 7.

src/cascade/executor/epiviz_runner.py
```diff
@@ -104,21 +106,46 @@ def add_omega_constraint(model_context, execution_context):
...
+        the age-standardized death rate is used to construct both
+        measurement data for mtall, with priors determined from
+        "measurement upper" and "measurement lower", but also a constraint
+        on omega, the underlying rate for other-cause mortality, so that
+        Dismod-AT will accept this as a given in the problem.
+
+     5. Create Average Integrand Cases, which are the list of
+        desired outputs from Dismod-AT to show in graphs in EpiViz-AT.
+
+     6. Construct all Fixed Effects.
```
nrhawkins: Could probably benefit from adding more detail here.

src/cascade/executor/epiviz_runner.py
```diff
@@ -104,21 +106,46 @@ def add_omega_constraint(model_context, execution_context):
...
+
+     7. Construct all Random Effects.
```
nrhawkins: and more detail here.

src/cascade/executor/epiviz_runner.py
```diff
@@ -173,6 +204,21 @@ def has_random_effects(model):
...
+
+     4. Using the Settings and the Execution Context, fill out the data and
+        any parameters for Dismod-AT, as described in
+        :ref:`build-model-from-epiviz-settings`.
+
+     5. Put that data into a file and run Dismod-AT on that file.
+    """
     ec = make_execution_context()
     settings = load_settings(ec, args.meid, args.mvid, args.settings_file)
     add_settings_to_execution_context(ec, settings)
```
nrhawkins: Just saying, it seems like we have a mix of "context.<attribute> = " and "add_something(context,...)"?

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+
+def _normalize_bundle_data(data):
+    """Normalize bundle columns, strip extra columns and index on `seq`
+    """
+    data = _normalize_measures(data)
+    data = _normalize_sex(data)
+
+    data = data.set_index("seq")
+
+    cols = ["measure", "mean", "sex", "sex_id", "standard_error", "age_start", "age_end", "year_start", "year_end", "location_id"]
```
nrhawkins: This line is long.  I can't see it all in the browser.  I don't see a scroll bar.

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+
+    weight_method = "constant"
+    MATHLOG.info(f"The set of weights for this bundle is {weight_method}.")
+    # Stick with year_start instead of time_start because that's what's in the
+    # bundle, so it's probably what modelers use. Would be nice to pair
+    # start with finish or begin with end.
+    return pd.DataFrame(
+        {
+            "measure": bundle_df["measure"],
+            "node_id": location_id,
```
nrhawkins: Ok, I have seen for the USA node_id = 98, and location_id = 102; i.e. they're not different names for the same number.
So, just wondering about that here?

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+    Convert bundle into an internal format. It removes the sex column and changes
+    location to node. It also adjusts for the demographic specification.
+    """
+    if "location_id" in bundle_df.columns:
+        location_id = bundle_df["location_id"]
+    else:
+        location_id = np.full(len(bundle_df), config.location_id, dtype=np.int)
+
+    # assume using demographic notation because this bundle uses it.
+    demographic_interval_specification = 0
```
nrhawkins: Can it be a 1 somewhere else, or are we planning to add a function call here?

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+    """Get bundle data with associated study covariate labels.
+
+    Args:
+        execution_context (ExecutionContext): The context within which to make the query
+        bundle_id (int): Bundle to load. Defaults to the bundle associated with the context
+        tier (int): Tier to load data from. Defaults to 3 (frozen data) but will also accept 2 (scratch space)
+
+    Returns:
+        A tuple of (bundle data, study covariate labels) where the bundle data is a pd.DataFrame and the labels are a
+        pd.DataFrame with an index aligned with bundle data and a column without ``x_`` for each study covariate.
```
nrhawkins: This requires knowledge that study covariate column names in bundles have names something like "x_<scov_name>",
which is maybe ok, but just pointing it out.

src/cascade/input_data/configuration/construct_study.py
```diff
@@ -0,0 +1,118 @@
...
+from collections.__init__ import defaultdict
+
+import numpy as np
+import pandas as pd
+
+from cascade.input_data import InputDataError
+from cascade.input_data.configuration.covariate_records import CovariateRecords
+from cascade.input_data.db.study_covariates import _get_study_covariates, \
+    covariate_ids_to_names
+from cascade.core.log import getLoggers
```
nrhawkins: Above in construct_country.py, we have this:  from cascade.core import getLoggers  
rather than: from cascade.core.log

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,74 @@
...
+ 2. Convert each record in the measurement bundle from its
+    GBD measure ID to a Dismod measure, according to the
+    `bundle id map <https://github.com/ihmeuw/cascade/blob/develop/src/cascade/input_data/configuration/id_map.py>`_.
+    If the incoming data uses incidence, instead of S-incidence
+    or T-incidence, reject it.
+
+ 3. Convert sex id from 1, 2, 3, 4 to male, female, both, or unspecified.
+
+ 4. Find the list of study covariates associated with this bundle,
+    both by covariate id and by name.
```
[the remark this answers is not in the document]
adolgert: The db has a list of covariates that came from the original, uploaded bundle. This gets that list and the values that were uploaded by the modeler.

docs/epiviz/overview.rst
adolgert: That's as much as I describe in this section, I suppose. This is a work in progress.

src/cascade/executor/epiviz_runner.py
adolgert: added a link to the definition of the model variables

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -0,0 +1,113 @@
...
+    except KeyError as ke:
+        raise RuntimeError(
+            f"The bundle data uses measure {str(ke)} which doesn't map "
+            f"to an integrand. The map is {gbd_measure_id_to_integrand}."
+        )
+    return data
+
+
+def _normalize_sex(data):
+    """Transform sex_ids from 1, 2, 3 to male, female, both.
```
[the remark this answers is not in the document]
adolgert: not this pr

src/cascade/input_data/configuration/construct_bundle.py
adolgert: haha

src/cascade/input_data/configuration/construct_bundle.py
adolgert: It gets a node_id from its order in the Dismod file, but the location id is the GBD assignment, and those go up to 50,000.

src/cascade/input_data/configuration/construct_bundle.py
adolgert: This is a bit of behavior that the modelers need to fix in order for the code to work correctly. This is a little yell that it's wrong.

src/cascade/input_data/configuration/construct_study.py
adolgert: Yeah, I let that slip. The logging is pulled up to core.

src/cascade/input_data/configuration/id_map.py
```diff
@@ -29,8 +29,8 @@
 14          15                    Cause-specific mortality rate mtspecific
 15          16                       Other cause mortality rate mtother
 16          17                               Case fatality rate
-17          18                                       Proportion
-18          19                                       Continuous
+17          18                                       Proportion Sincidence
```
alec-deason: Really? This is for the single parameter models?

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,76 @@
...
+
+Introduction
+------------
+EpiViz-AT is a web page on http://epimodeling-web-d01.ihme.washington.edu/at
+or http://epimodeling-web-p01.ihme.washington.edu/at. After you fill out
+that web page and hit submit, the web server runs a program on the cluster.
+We call that program ``epiviz_runner``. It is installed on the IHME cluster,
+and the code itself is stored at https://github.com/ihmeuw/cascade.
+The ``epiviz_runner``, itself, is at
+https://github.com/ihmeuw/cascade/blob/develop/src/cascade/executor/epiviz_runner.py.
```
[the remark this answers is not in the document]
adolgert: I'm questioning this whole section of docs, really. I needed to write this in order to get my head straight on how to do study covariates. For instance, I knew the bundle had them but forgot that all of the measurements, including the mtall that we add to it, need to have these same study covariates. I do want one reference so that there is a breadcrumb. Maybe I can make it a "ref" though...

examples/simple_load_from_epiviz.py
```diff
@@ -8,7 +8,9 @@
 from cascade.input_data.db.configuration import settings_json_from_epiviz
 from cascade.executor.no_covariate_main import bundle_to_observations
 from cascade.input_data.configuration.form import Configuration
-from cascade.input_data.db.bundle import bundle_with_study_covariates, freeze_bundle
+from cascade.input_data.configuration.construct_bundle import \
```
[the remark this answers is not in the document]
adolgert: Yes, the darned editor is doing this. I need to fix everywhere.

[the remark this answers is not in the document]
adolgert: I fixed them everywhere.

src/cascade/executor/epiviz_runner.py
adolgert: I'm weighing what should be close to the code and what should be far. I'll argue for these two docstrings on the grounds that they are the two highest-level and most important functions in the whole application.

src/cascade/executor/epiviz_runner.py
```diff
@@ -95,30 +98,59 @@ def add_omega_constraint(model_context, execution_context):
...
+    """
     model_context = initial_context_from_epiviz(settings)
 
-    fixed_effects_from_epiviz(model_context, execution_context, settings)
-    random_effects_from_epiviz(model_context, settings)
-
     freeze_bundle(execution_context, execution_context.parameters.bundle_id)
     load_csmr_to_t3(execution_context)
     load_asdr_to_t3(execution_context)
+    execution_context.parameters.tier = 3
```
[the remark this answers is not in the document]
adolgert: Does it help if tiers are going away? I do know what a tier is. It's one of the few organizing principles I've seen out of central comp.

src/cascade/executor/epiviz_runner.py
adolgert: The definition of fixed and random effects isn't strong in statistics literature, and it's not strong here. I'm pushing for representations in the Model object that say what the moving parts are (covariate, multiplier, priors). The moment we call them fixed and random should be when changing to the dismod file. Maybe this distinction is better put lower in the code. It's good that it's at the end of the translation process. We're not far off in terms of representation.

src/cascade/executor/epiviz_runner.py
adolgert: I really appreciate Natalie's observation that this variation in call semantics causes surprise. We should make the API really, really boring. I'm not up for resolving it in this PR.

src/cascade/input_data/configuration/construct_bundle.py
adolgert: It's the number of years to add. Maybe I could change the variable name? This part of the code is a terrible FIXME moment b/c we don't trust the data at all. This should be reading a column of True or False and then adding 1 or 0.

src/cascade/input_data/configuration/construct_bundle.py
adolgert: fixed

src/cascade/input_data/configuration/id_map.py
adolgert: That's where the data goes. It's for the single-parameter models, yes. This came out of a modeler meeting. What we then do with that data is another matter.

docs/epiviz/overview.rst
```diff
@@ -0,0 +1,76 @@
...
+=========================
+
+.. _epiviz-overview-introduction:
+
+Introduction
+------------
+EpiViz-AT is a web page on http://epimodeling-web-d01.ihme.washington.edu/at
+or http://epimodeling-web-p01.ihme.washington.edu/at. After you fill out
+that web page and hit submit, the web server runs a program on the cluster.
+We call that program ``epiviz_runner``. It is installed on the IHME cluster,
```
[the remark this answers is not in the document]
adolgert: fixed

src/cascade/executor/epiviz_runner.py
```diff
@@ -79,6 +80,8 @@ def add_mortality_data(model_context, execution_context):
     )
     csmr["measure"] = "mtspecific"
     csmr = csmr.rename(columns={"location_id": "node_id"})
+    # Ensure that the csmr index doesn't overlap with the existing seq values in the observations.
+    csmr.index = csmr.index.values + model_context.input_data.observations.index.max() + 1
```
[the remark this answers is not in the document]
adolgert: This is the main comment... and you're right. This was uncomfortable, which is a sign. I think I need the seq as an index for joins. It should be a column, though, so it can't be lost. And we need to keep `hold_out` as a column too instead of using `constraints`. Let me make a commit and then see what I can do.

src/cascade/executor/epiviz_runner.py
alec-deason: I question any comment that obscures what is actually happening. If we're going to have high level step-by-steps like this in the code then I think they go at module level? At least they are out of the way.

src/cascade/executor/epiviz_runner.py
alec-deason: If we want one semantic, then it has to be setter functions I think.

src/cascade/input_data/configuration/construct_bundle.py
alec-deason: Oh, that was not clear to me. I think a different name would help `demographic_interval_offset` or something maybe.

PR #69, opened 2018-10-24
more informative names for priors and smooths in the database

alec-deason:


adolgert:
The functionality I'm hoping for is, I think, here. If you name the smooth, then it prepends its name onto the priors it saves. Mostly this is for Greg, though, so whatever works for him.

PR #72, opened 2018-10-26
the central comp save results changed for the worse but this sets yea…

adolgert:
This fixes save_results so that it handles 1) x_<number> instead of x_sex, like we expected and 2) save_results_at now takes gbd_round_id. Not just that, but it puts the expected years into an initializer that's a list, which isn't good practice. Anyhow, we set the IDs ourselves.

Merging this makes Dismod work for again the modelers. Who wants to do the honors?

### DIFF
--- a/src/cascade/saver/db/save_model_results_to_db.py
+++ b/src/cascade/saver/db/save_model_results_to_db.py
@@ -90,9 +90,12 @@ def _normalize_draws_df(draws_df, execution_context):
     node_to_location = {r.node_id: r.c_location_id for _, r in node_table.iterrows()}
 
     draws["location_id"] = draws.node_id.apply(lambda nid: node_to_location[nid])
-
-    draws["sex_id"] = draws.x_sex.apply(lambda x: {-0.5: 2, 0.5: 1}[x])
-    return draws.drop(["node_id", "weight_id", "x_sex"], "columns")
+    covariate_table = execution_context.dismodfile.covariate
+    sex_index = int(covariate_table[covariate_table.covariate_name == "sex"].covariate_id)
+    draws["sex_id"] = draws[f"x_{sex_index}"].apply(lambda x: {-0.5: 2, 0.5: 1}[x])
+    # Remove covariates from draws to upload.
+    to_drop = ["node_id", "weight_id"] + [str(cov_col) for cov_col in draws.columns if cov_col.startswith("x_")]
+    return draws.drop(to_drop, "columns")
 
 
 def _write_temp_draws_file_and_upload_model_results(draws_df, execution_context):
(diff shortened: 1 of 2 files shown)


src/cascade/saver/db/save_model_results_to_db.py
```diff
@@ -90,9 +90,12 @@ def _normalize_draws_df(draws_df, execution_context):
...
 
     draws["location_id"] = draws.node_id.apply(lambda nid: node_to_location[nid])
-
-    draws["sex_id"] = draws.x_sex.apply(lambda x: {-0.5: 2, 0.5: 1}[x])
-    return draws.drop(["node_id", "weight_id", "x_sex"], "columns")
+    covariate_table = execution_context.dismodfile.covariate
+    sex_index = int(covariate_table[covariate_table.covariate_name == "sex"].covariate_id)
+    draws["sex_id"] = draws[f"x_{sex_index}"].apply(lambda x: {-0.5: 2, 0.5: 1}[x])
+    # Remove covariates from draws to upload.
+    to_drop = ["node_id", "weight_id"] + [str(cov_col) for cov_col in draws.columns if cov_col.startswith("x_")]
```
alec-deason: Does it make more sense to white list columns?

src/cascade/saver/db/save_model_results_to_db.py
```diff
@@ -90,9 +90,12 @@ def _normalize_draws_df(draws_df, execution_context):
     node_to_location = {r.node_id: r.c_location_id for _, r in node_table.iterrows()}
 
     draws["location_id"] = draws.node_id.apply(lambda nid: node_to_location[nid])
-
-    draws["sex_id"] = draws.x_sex.apply(lambda x: {-0.5: 2, 0.5: 1}[x])
-    return draws.drop(["node_id", "weight_id", "x_sex"], "columns")
+    covariate_table = execution_context.dismodfile.covariate
+    sex_index = int(covariate_table[covariate_table.covariate_name == "sex"].covariate_id)
```
alec-deason: What happens if there are multiple sex covariates? Or are we assuming that is sufficiently impossible?

src/cascade/saver/db/save_model_results_to_db.py
```diff
@@ -90,9 +90,12 @@ def _normalize_draws_df(draws_df, execution_context):
...
+    draws["sex_id"] = draws[f"x_{sex_index}"].apply(lambda x: {-0.5: 2, 0.5: 1}[x])
```
nrhawkins: Does this mean that there are no 0's in the sex values?

src/cascade/saver/db/save_model_results_to_db.py
adolgert: It was a blacklist. I'm not sure why. Need to have a longer conversation with Natalie. This is why we try to make the first choice correct, because the first choice lives a long time.

src/cascade/saver/db/save_model_results_to_db.py
adolgert: Every covariate is saved only once because of the machinery I made. So if you have sex used by two covariate multipliers, then they use the same column. If sex is transformed, it becomes sex_log or sex_squared, which then gets its own column.

src/cascade/saver/db/save_model_results_to_db.py
adolgert: Oh, dear. This is wrong. And it needs to be near. To be fair, I didn't modify that part of the code. I changed only the name of the variable.

PR #73, opened 2018-10-29
Bugfix/loglines

adolgert:
* Added <br /> to end of mathlog lines.
* Changed Dismod-AT stderr to warning level b/c it prints lines to stderr.
* Changed Dismod-AT messages to use a real function instead of a lambda so they print better in the log.

### DIFF
--- a/src/cascade/executor/argument_parser.py
+++ b/src/cascade/executor/argument_parser.py
@@ -203,7 +203,8 @@ class BaseArgumentParser(ArgumentParser):
         log_file = math_log_dir / "log.log"
         try:
             append_to_math_log = "a"
-            math_handler = logging.FileHandler(str(log_file), append_to_math_log)
+            windows_newlines = "\r\n"
+            math_handler = logging.StreamHandler(log_file.open(append_to_math_log, newline=windows_newlines))
         except (OSError, PermissionError) as mhe:
             logging.warning(f"Could not write to math log at {log_file} even though "
                             f"directory {math_log_dir} exists: {mhe}")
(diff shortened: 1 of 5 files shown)


src/cascade/executor/argument_parser.py
```diff
@@ -208,7 +216,8 @@ def _logging_configure_mathlog(mvid, epiviz_log_dir):
             logging.warning(f"Could not write to math log at {log_file} even though "
                             f"directory {math_log_dir} exists: {mhe}")
             return
-        fmt = "%(levelname)s %(asctime)s %(funcName)s: %(message)s"
+        # The br is an HTML tag to add a line break.
+        fmt = "%(levelname)s %(asctime)s %(funcName)s: %(message)s<br />"
```
alec-deason: This implies that the math log is _only_ to be shown in the GUI since having random HTML in it makes it less readable in other contexts. Be nice to have the viz people handle this instead but that's fine.

src/cascade/executor/dismod_runner.py
```diff
@@ -167,6 +167,20 @@ def run_and_watch(command, single_use_machine, poll_time):
...
+    Otherwise, they show <lambda> as the function name.
+    """
+    MATHLOG.info(text)
+
+
+def dismod_report_stderr(text):
+    """This ensures MATHLOG messages have a function name in the log.
+    Otherwise, they show <lambda> as the function name.
+    """
+    MATHLOG.warning(text)
```
nrhawkins: down below, this used to be MATHLOG.error, rather than warning.  Is that intentional?

PR #76, opened 2018-11-02
Feature/mortality sex covariate

adolgert:
This subselects for a particular sex in the mortality and omega constraint.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -99,6 +99,7 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     if model_context.input_data.times:  # The times are a set so can be tested this way.
         min_time = np.min(list(model_context.input_data.times))  # noqa: F841
         max_time = np.max(list(model_context.input_data.times))  # noqa: F841
+        # The % 5 is to exclude annual data points.
         asdr = asdr.query("time_lower >= @min_time and time_upper <= @max_time and time_lower % 5 == 0")
     model_context.rates.omega.parent_smooth = build_constraint(asdr)
     MATHLOG.debug(f"Add {asdr.shape[0]} omega constraints from age-standardized death rate data.")
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -67,36 +67,38 @@ def meas_bounds_to_stdev(df):
     return df.drop(["meas_lower", "meas_upper"], axis=1)
 
 
-def add_mortality_data(model_context, execution_context):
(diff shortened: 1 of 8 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -67,40 +67,45 @@ def meas_bounds_to_stdev(df):
...
     asdr["measure"] = "mtall"
     asdr = asdr.rename(columns={"location_id": "node_id"})
-    min_time = np.min(list(model_context.input_data.times))  # noqa: F841
-    max_time = np.max(list(model_context.input_data.times))  # noqa: F841
-    asdr = asdr.query("time_lower >= @min_time and time_upper <= @max_time and time_lower % 5 == 0")
+    asdr = asdr.query(f"sex_id == @sex_id")
+    if model_context.input_data.times:  # The times are a set so can be tested this way.
+        min_time = np.min(list(model_context.input_data.times))  # noqa: F841
+        max_time = np.max(list(model_context.input_data.times))  # noqa: F841
+        asdr = asdr.query("time_lower >= @min_time and time_upper <= @max_time and time_lower % 5 == 0")
```
alec-deason: I know that this `%5` is to exclude the weird annual data points but that deserves a comment because it ain't obvious.

src/cascade/executor/epiviz_runner.py
```diff
@@ -156,8 +161,9 @@ def model_context_from_settings(execution_context, settings):
         MATHLOG.warning(f"removing {remove_cnt} rows from bundle where standard_error == 0.0")
         model_context.input_data.observations = model_context.input_data.observations[mask]
 
-    add_mortality_data(model_context, execution_context)
-    add_omega_constraint(model_context, execution_context)
+    settings.model.drill_sex = 1
```
alec-deason: This is stray debugging code?

src/cascade/input_data/db/mortality.py
```diff
@@ -7,14 +7,19 @@
...
+    """
     model_version_id = execution_context.parameters.model_version_id
 
     query = """
      SELECT
                 model_version_dismod_id as data_id,
                 location_id,
                 year_start as time_lower,
-                year_end as time_upper,
+                year_end + 1 as time_upper,
```
alec-deason: Awesome. You did exactly the same thing I did. Character for character. This feels wrong to me because it's stuck directly in the SQL. But doing it outside the SQL seems silly. I don't know. I guess it doesn't matter.

src/cascade/executor/epiviz_runner.py
adolgert: yes! totally a mistake. thank you so much.

PR #82, opened 2018-11-05
fixed all flake8 errors so we can add to Travis

adolgert:
I'd like to add flake8 to Travis, so this fixes the current errors.


setup.cfg
```diff
@@ -1,7 +1,7 @@
 [flake8]
 max-line-length = 120
 # E731 do not assign a lambda expression, use a def
-ignore = E731
+ignore = E731,W503,W504
```
nrhawkins: Should we add explanations for the 2 W's, like for E731?

src/cascade/input_data/configuration/form.py
```diff
@@ -49,9 +50,8 @@ def _full_form_validation(self, root):
             lower = self.min
             upper = self.max
             mean = self.mean
-            if mean is None:
-                if np.isinf(lower) or np.isinf(upper):
-                    mean = max(lower, 0)
+            if mean is None and np.isinf(lower) or np.isinf(upper):
```
nrhawkins: Do we need parens here?  if mean is None and (np.isinf(lower) or np.isinf(upper))

Say, mean is 3, and upper is np.inf and lower is 1, then mean will be set to 1.  
Can lower and upper be None?  doesn't look like np.isinf() will like it, if so?

PR #83, opened 2018-11-05
Feature/use default nu and eta

alec-deason:
This is a real simple one that just brings in the global eta and nu parameters from the epiviz configs. Also fills in some gaps in the tests


src/cascade/input_data/configuration/form.py
```diff
@@ -63,11 +73,11 @@ def _full_form_validation(self, root):  # noqa: C901 too complex
                 eta = self.eta
 
             if self.density == "uniform":
-                self.prior_object = priors.Uniform(lower, upper, mean)
+                self.prior_object = priors.Uniform(lower, upper, mean, eta=eta)
```
adolgert: I don't think eta affects uniform, does it? eta affects the prior for log-Gaussian, log-Laplace, and log-Students.

tests/saver/db/test_save_model_results_to_db.py
```diff
@@ -152,13 +151,17 @@ def test_write_temp_draws_file_and_upload_model_results_no_hdf_no_sr_call(
     draws_df, execution_context, fake_write_hdf
 ):
 
-    model_version_id = _write_temp_draws_file_and_upload_model_results(draws_df, execution_context, saver=save_results_fake)
+    model_version_id = _write_temp_draws_file_and_upload_model_results(draws_df,
```
adolgert: I'm less fond of hanging arguments near the edge. Have you considered either a) doing a return right after the left-paren and lumping them in there in one line or b) doing a return after the left-paren and hanging them there? This hanging stuff means that, when we change anything in the first line, we have to reformat the rest of the lines, and, while there's a keystroke for that, it seems a waste. It's also visually jarring when successive calls have hanging values at different columns, so you get a river-winding look.

src/cascade/input_data/configuration/form.py
adolgert: I don't think there is a log-uniform. So I don't think this one gets the eta.

alec-deason: Brad's docs seem to indicate that it effects any prior, but I don't really follow what he's saying:

https://bradbell.github.io/dismod_at/doc/prior_table.htm

src/cascade/input_data/configuration/form.py
```diff
@@ -53,7 +53,17 @@ def _full_form_validation(self, root):  # noqa: C901 too complex
             if mean is None and np.isinf(lower) or np.isinf(upper):
                 mean = max(lower, 0)
             std = self.std
-            nu = self.nu
+
+            if self.nu is None:
```
nrhawkins: Are there any comments around these parts about what "nu" is?  or "why" nu exists?

If not, if there is a short way to add a little bg to this chunk, it could be nice, I think.

src/cascade/input_data/configuration/form.py
```diff
@@ -53,7 +53,17 @@ def _full_form_validation(self, root):  # noqa: C901 too complex
...
+                if self.density == "students" and not root.is_field_unset("students_dof"):
+                    nu = root.students_dof.priors
+                elif self.density == "log_students" and not root.is_field_unset("log_students_dof"):
+                    nu = root.log_students_dof.priors
+                else:
+                    nu = None
+            else:
+                nu = self.nu
+
             if self.eta is None:
```
nrhawkins: similarly, for eta.  eta (maybe) helps prevent getting a log value of -inf, by keeping the value a little bit to the right of zero.

tests/core/form/test_abstract_form.py
```diff
@@ -13,7 +13,7 @@ def _validate_and_normalize(self, instance, value):
             return new_value, None
 
     class MyInnerForm(Form):
-        my_inner_field = IntField()
+        my_inner_field = IntField(display="My Inner Field!!")
```
nrhawkins: I like the enthusiasm!!

tests/input_data/configuration/test_form.py
```diff
@@ -1,29 +1,102 @@
...
+    }
+
+    f = DummyForm(
+        {
+            "log_students_dof": {"priors": 5, "data": 5},
+            "prior": {"prior_type": "value", "density": "students", "mean": 0, "std": 0.1},
+        }
+    )
+    assert set(f.validate_and_normalize()) == {
+        ("prior", "prior", "Parameters incompatible with density 'students': Nu must be positive: nu=None")
```
nrhawkins: is the density here "log_students_dof" (line 96), or "students" (line 97)?

What I am saying is that in line 101, the incompatibility would be with the "log students" not with "students" as it says? given that in the examples above, it appears that nu does not have to be supplied for "students"?

src/cascade/input_data/configuration/form.py
adolgert: There are two parts to the section about eta. The first says it applies only to log-transformed values. The second is confusing, but it looks to me like an offset for those same log-transformed values. So I think it's not on the Uniform distribution.

alec-deason: Ok. I'll pull them from the inappropriate distributions.

tests/input_data/configuration/test_form.py
alec-deason: The error is that there is a students-t distribution which did not have it's own nu and there was a global nu for log students-t but not one for students-t.

students-t and log students-t must always have a nu, but it can be supplied by either type specific global settings or local settings, which is what this test exercises.

I'll add some comments because it isn't clear.

tests/saver/db/test_save_model_results_to_db.py
alec-deason: Sure

src/cascade/input_data/configuration/form.py
alec-deason: I don't think this is the place for that kind of documentation. I'll make a note in the code where the parameters are actually defined that points to Brad's documentation of how they are used.

alec-deason:
Ok, I think I have addressed everybody's suggestions.

PR #85, opened 2018-11-06
Write csv files so modelers can compare AT and ODE save results data

nrhawkins:
This is the second PR for this task.  Database and table are now offered as arguments to the program, so that we can pin-point the source of the data.  In the previous version, one db and table was checked for AT and a separate db and table for ODE.  When a modeler ran the program, no data was found, because though the data exists, it did not exist in the one place checked for each model type.

### DIFF
--- a/src/cascade/executor/model_results_main.py
+++ b/src/cascade/executor/model_results_main.py
@@ -15,28 +15,42 @@ from cascade.core.db import connection
 CODELOG = logging.getLogger(__name__)
 
 
-dbs = ["epi-dev":"epi-test",
-           "epi-prod":"epi",
-           "at-dev":"dismod-at-dev",
-           "at-prod":"dismod-at-prod"]
+dbs = {"epi-dev": "epi-test",
+       "epi-prod": "epi",
+       "at-dev": "dismod-at-dev",
+       "at-prod": "dismod-at-prod"}
 
-tables = {"fit":"epi.model_estimate_fit", "final":"epi.model_estimate_final"}
+dbs_easy_names = {"epi-test": "epi-dev",
+                  "epi": "epi-prod",
+                  "dismod-at-dev": "at-dev",
+                  "dismod-at-prod": "at-prod"}
+
+tables = {"fit": "epi.model_estimate_fit", "final": "epi.model_estimate_final"}
+
+tables_easy_names = {"epi.model_estimate_fit": "fit", "epi.model_estimate_final": "final"}
 
 
-def _check_database(db, table):
-    """
+def _retrieve_from_database(model_version_id, db, table):
+    """Connect to the database and retrieve the results.
+
+    Args:
+        db (str): nickname of the IHME database, as used in a .odbc.ini file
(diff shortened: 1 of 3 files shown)


src/cascade/executor/model_results_main.py
```diff
@@ -32,27 +73,35 @@ def _get_model_results(model_version_id, model_type):
...
-
-    with connection(database=database) as c:
-        model_results_data = pd.read_sql(query, c, params={"model_version_id": model_version_id})
-        CODELOG.debug(f"""Downloaded {len(model_results_data)} lines of dismod {model_type}
-                      data for model_version_id {model_version_id} from '{database}'""")
-
-    return model_results_data
+        # check all, if find more than one report it, else return one
+        model_results = None
+        model_results_empty = None
```
adolgert: This is new to me. In order to tell the reader that the dataframe is empty, the code assigns it to a variable with the name "empty". Think of this algorithm in terms of the minimum number of variables to define its state. There are two questions: what's available, and, if only one is available, what are its values? The function commits to retrieving everything it finds, so go with that. In short: let model_results be a list. Then count at the end.

nrhawkins: Thanks for commenting on this, because I didn't really like my final answer, but.
Here's what I was thinking.  If I keep non-empty's in model_results, then I can return the one non-empty one.  I can't, however, return an empty one (which would be something that has all the column names, but no data rows, which gets returned from the db retrieval).  If I keep empty's and non-empty's in model results, and there is only one non-empty in the list, I need to know which one it is.

nrhawkins: If an error is immediately raised to indicate multiple results, we can't report to them where all the results are, to help them choose one.

src/cascade/executor/model_results_main.py
```diff
@@ -32,27 +73,35 @@ def _get_model_results(model_version_id, model_type):
...
+        for db, table in it.product(dbs.values(), tables.values()):
+            model_results_retrieved = _retrieve_from_database(model_version_id, db, table)
+            if not model_results_retrieved.empty:
+                model_results_db_and_table.append((dbs_easy_names[db], tables_easy_names[table]))
+                model_results = model_results_retrieved
+            else:
+                model_results_empty = model_results_retrieved
+        if len(model_results_db_and_table) > 1:
+            # exit and report
+            sys.exit(f"""Found {model_version_id} in multiple (database, table) locations
```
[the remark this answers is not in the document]
nrhawkins: Ok, sure, thanks for this!  I used to use parens with if's...  in the beginning....  :-)

src/cascade/executor/model_results_main.py
```diff
@@ -14,7 +15,47 @@
 CODELOG = logging.getLogger(__name__)
 
 
-def _get_model_results(model_version_id, model_type):
+dbs = {"epi-dev": "epi-test",
```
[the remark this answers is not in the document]
nrhawkins: ok, sure.

PR #88, opened 2018-11-06
Feature/birth prev as prev at age zero

alec-deason:
This adds age group 164 as a special case if the modeler has selected birth prevalence.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -25,9 +25,11 @@ from cascade.input_data.configuration.builder import (
     initial_context_from_epiviz,
     fixed_effects_from_epiviz,
     random_effects_from_epiviz,
-    build_constraint)
+    build_constraint,
+)
 
 from cascade.core import getLoggers
+
 CODELOG, MATHLOG = getLoggers(__name__)
 
 
@@ -80,8 +82,9 @@ def add_mortality_data(model_context, execution_context, sex_id):
     csmr = csmr.rename(columns={"location_id": "node_id"})
     csmr = csmr.query(f"sex_id == @sex_id")
     MATHLOG.debug(f"Creating a set of {csmr.shape[0]} mtspecific observations from IHME CSMR database.")
-    model_context.input_data.observations = pd.concat([model_context.input_data.observations, csmr],
-                                                      ignore_index=True, sort=True)
+    model_context.input_data.observations = pd.concat(
+        [model_context.input_data.observations, csmr], ignore_index=True, sort=True
+    )
 
 
 def add_omega_constraint(model_context, execution_context, sex_id):
@@ -105,8 +108,9 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     MATHLOG.debug(f"Add {asdr.shape[0]} omega constraints from age-standardized death rate data.")
 
     mask = model_context.input_data.observations.measure == "mtall"
-    model_context.input_data.constraints = pd.concat([model_context.input_data.observations[mask], asdr],
-                                                     ignore_index=True, sort=True)
+    model_context.input_data.constraints = pd.concat(
+        [model_context.input_data.observations[mask], asdr], ignore_index=True, sort=True
+    )
     model_context.input_data.observations = model_context.input_data.observations[~mask]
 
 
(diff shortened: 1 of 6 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -230,6 +233,10 @@ def main(args):
     """
     ec = make_execution_context()
     settings = load_settings(ec, args.meid, args.mvid, args.settings_file)
+
+    if settings.model.drill != "drill":
+        raise NotImplementedError("Only 'drill' mode is currently supported")
```
adolgert: Ah, right. Nice.

src/cascade/model/integrands.py
```diff
@@ -7,20 +7,24 @@
 CODELOG, MATHLOG = getLoggers(__name__)
 
 
-def make_average_integrand_cases_from_gbd(execution_context, sexes):
+def make_average_integrand_cases_from_gbd(execution_context, sexes, include_birth_prevalence=False):
+    """
+    """
```
adolgert: Curious empty comment block?

src/cascade/model/integrands.py
```diff
@@ -7,20 +7,24 @@
...
     gbd_age_groups = get_age_groups(execution_context)
     age_ranges = [(r.age_group_years_start, r.age_group_years_end) for _, r in gbd_age_groups.iterrows()]
     time_ranges = [(y, y) for y in get_years(execution_context)]
 
+    # Assuming using the first set of weights, which is constant.
+    weight_id = 0
```
adolgert: We do not have our act together regarding weights. This is not the PR in which to fix that.

src/cascade/saver/db/save_model_results_to_db.py
```diff
@@ -64,8 +64,23 @@ def _normalize_draws_df(draws_df, execution_context):
...
+    # age group set and which the model may optionally produce for prevalence (and no other
+    # measure). I'm hard coding it in here but it really should be plumbed through better.
+    age_groups = age_groups.append([{
+        "age_group_id": 164,
+        "age_group_years_start": 0,
+        "age_group_years_end": 0,
+    }])
+
+    # TODO: This assumes there will never be two different age groups with the same
+    # upper range. Be wary
```
adolgert: I'm OK with that restriction if these are draws b/c we make the draws. They won't have age ranges that are summaries, combinations of ages together.

tests/model/test_integrands.py
```diff
@@ -32,3 +32,34 @@ def test_make_average_integrand_cases_from_gbd(mocker):
...
+        for (time_lower, time_upper) in {(1990, 1990), (1995, 1995), (2000, 2000)}:
+            for integrand in IntegrandEnum:
+                assert len(
+                    average_integrand_cases.query(
+                        "age_lower == @age_lower and age_upper == @age_upper "
+                        "and time_lower == @time_lower and time_upper == @time_upper "
+                        "and integrand_name == @integrand.name"
+                    )
+                    == 1
+                )
```
adolgert: We need to reshuffle at a larger scale in order to improve testability here.

src/cascade/model/integrands.py
alec-deason: Well, crap. I started writing that and then I guess I got distracted.

src/cascade/model/integrands.py
alec-deason: Yep, this is the third of fourth time I've carried this forward without really understanding it.

src/cascade/saver/db/save_model_results_to_db.py
alec-deason: I am conscious of it because I just add the birth age group which shares it's starting age with the neo-natal group and that broke some assumptions.

tests/model/test_integrands.py
alec-deason: I agree that this isn't a great test but I'm not sure I understand what you mean. Are you just saying that testing over these single, static examples isn't as good as testing over a space of possible examples?

src/cascade/executor/epiviz_runner.py
alec-deason: Trying to cinch the validation and checks a little tighter every time I touch something.

PR #89, opened 2018-11-07
Feature/constraints join observations

adolgert:
This gets rid of the input_data.constraints completely. Replaces it with a hold_out column on the observations. This makes covariates work for Greg.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -109,7 +109,7 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     MATHLOG.debug(f"Add {asdr.shape[0]} omega constraints from age-standardized death rate data.")
 
     observations = model_context.input_data.observations
-    observations[observations.measure == "mtall"][["hold_out"]] = 1
+    observations.loc[observations.measure == "mtall", "hold_out"] = 1
     asdr = asdr.assign(hold_out=1)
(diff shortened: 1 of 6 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -107,11 +108,10 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     model_context.rates.omega.parent_smooth = build_constraint(asdr)
     MATHLOG.debug(f"Add {asdr.shape[0]} omega constraints from age-standardized death rate data.")
 
-    mask = model_context.input_data.observations.measure == "mtall"
-    model_context.input_data.constraints = pd.concat(
-        [model_context.input_data.observations[mask], asdr], ignore_index=True, sort=True
-    )
-    model_context.input_data.observations = model_context.input_data.observations[~mask]
+    observations = model_context.input_data.observations
+    observations[observations.measure == "mtall"][["hold_out"]] = 1
```
nrhawkins: Is observations a pd.DataFrame()?   I was just trying something like this on the side, and I am getting an error: 
test_nh.py:9: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.

I am just sharing, so you can check.

adolgert: Thanks so much for catching this. It's a classic mistake when using Pandas.

PR #90, opened 2018-11-08
Exclude outliers from bundles

alec-deason:
This is the source of one of the data problems Greg had in the model he showed Theo. Tom agreed that ignoring these outliers is the right approach but I wanted to make sure you were aware that it's happening.

### DIFF
--- a/src/cascade/input_data/db/bundle.py
+++ b/src/cascade/input_data/db/bundle.py
@@ -57,7 +57,7 @@ def _get_bundle_id(execution_context):
         return bundle_ids[0][0]
 
 
-def _get_bundle_data(execution_context, bundle_id, tier=3):
+def _get_bundle_data(execution_context, bundle_id, tier=3, exclude_outliers=True):
     """
     Downloads the tier 2 or 3 data for the bundle associated with the current
     model_version_id.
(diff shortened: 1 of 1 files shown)


adolgert:
My one question is whether we want to exclude them or mark them as hold_out. If it's mark as hold out, then read them as outlier_type_id, and change them to hold_out=1 in the normalize() function.

alec-deason:
So, the conversation we (Greg, Tom and I) had was that these should be completely suppressed. Tom, as an outsider, expressed surprise that we were using "holdout" for things other than cross validation which I tend to agree is strange. Maybe it's worth asking the modelers if there is ever any reason why they would want to see these points inside the model, it could only be for debugging.

adolgert:
OK. The hold out gives you a way to see residuals on things that don't go into the model. That's what it gives you. I could see modelers wanting to see the outliers as hollow dots on EpiViz-AT.
Makes sense to respect the outlier flag now and ask if they want different behavior.

alec-deason:
Oh, wait. You are right. I missed this point. The code as written removes the outliered points from the tier3 bundle which means you won't see them at all in the GUI. That's not what we want. We only want to remove the points from the bundle we use internally. Let me tweak.

alec-deason:
Also my first pass at this didn't handle the study covariates correctly.

alec-deason:
No, covariates are fine. You're already doing a merge that allows for sparsity on both sides.

alec-deason:
Ok. That's more complicated but I think it does what we actually want.

PR #91, opened 2018-11-09
Acceptance tests around mortality data

alec-deason:
One outstanding question I have here is how to think about the relationship between these tests and the database. They depend on building a fresh tier3 and so must delete the tier3 data for the mvid they test and reupload the data. I've chosen a mvid that no one cares about but it may disappear at some point and the fact the test is dramatically modifying a shared database at all makes me nervous. I'm not sure how to do better at this point.

### DIFF
--- a/acceptance/test_mortality.py
+++ b/acceptance/test_mortality.py
@@ -1,0 +1,108 @@
+from subprocess import run
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+import numpy as np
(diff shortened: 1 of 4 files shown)


alec-deason:
There's nothing in here about the bundle, though we could add a test for that. This is to prove that the mortality data we pull from GBD is the same mortality data that ends up in the dismod db. Because there was a question about that.

PR #92, opened 2018-11-13
Naive EMR from Prevalence and CSMR calculation

alec-deason:
This introduces code to estimate EMR from modeler supplied Prevalence data and GBD supplied CSMR data. It uses a naive interpolation over CSMR which makes both Drew and Greg sad but it's what's available today.

Currently this code is not used. My next PR will introduce a concept of "policies" the first of which will be the ability to select whether or not to use this method.


src/cascade/input_data/db/demographics.py
```diff
@@ -30,20 +31,38 @@ def get_locations(execution_context):
...
+        with_group = with_group.drop(columns="age_group_id")
     return with_group.rename(columns={"age_group_years_start": "age_lower", "age_group_years_end": "age_upper"})
 
 
-def get_all_age_spans(execution_context):
+def get_mean_years(execution_context):
+    """
+    Get the mean year estimate for age groups. This is the mean age beyond the
+    start of the age group for people in a given age group.
+
```
nrhawkins: Maybe add a numerical example here.  I am not familiar with the get_life_table, though it sounds like a nice one!

Are you saying in the comments from the other PR, that if an age group is 5-10, this might return 2.5?

get_years_from_lower_age_to_mean_age

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+
+CODELOG, MATHLOG = getLoggers(__name__)
+
+
+def _prepare_prevalence(observations):
+    prevalence = observations.query("measure == 'prevalence' and sex_id in [1, 2, 3]")
+
+    zero_means = prevalence["mean"] == 0
+    if np.any(zero_means):
+        not_quite_zero = 1e-6
```
nrhawkins: If I were to do this, would this be ok, or would I be asked to create a module constant?  

sometimes, looks like we use 1e-9, here 1e-6.

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+    return csmr.drop(columns=["time_lower", "time_upper"])
+
+
+def _collapse_ages_unweighted(csmr):
+    csmr["age"] = (csmr["age_lower"] + csmr["age_upper"]) / 2
+    return csmr
+
+
+def _collapse_ages_weighted(execution_context, csmr):
+    mean_years = get_mean_years(execution_context)
```
nrhawkins: Here, I am starting to think - how is this weighting done?  the secret, I think, is in the awesome get_life_table.  So, maybe add "weighting" to the comments for get_mean_years() docstring, or maybe add a short comment to this fcn?

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+        return pd.Series({"mean": float(emr_mean), "standard_error": float(emr_stderr)})
+
+    return emr.merge(prevalence.apply(emr_mean, "columns"), left_index=True, right_index=True)
+
+
+def add_emr_from_prevalence(model_context, execution_context):
+    r"""Estimate excess mortality from the supplied prevalence measurements and
+    cause specific mortality estimates from GBD using the formula:
+    :math:`\chi=\frac{P}{{}_nm_x^c}
```
nrhawkins: I am trying to read this, and I am having trouble...

src/cascade/stats/utils.py
```diff
@@ -0,0 +1,26 @@
...
+
+def meas_bounds_to_stdev(df):
+    r"""
+    Given data that includes a measurement upper bound and measurement lower
+    bound, assume those are 95% confidence intervals. Convert them to
+    standard error using:
+
+    .. math::
+
+        \mbox{stderr} = \frac{\mbox{upper} - \mbox{lower}}{2 1.96}
```
nrhawkins: I need to do some work, I can't quite read these.  Does that mean 2*1.96? (which is what I see below)

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+
+def _emr_from_sex_and_node_specific_csmr_and_prevalence(csmr, prevalence):
+    mean_interp, stderr_interp = _make_interpolators(csmr)
+
+    emr = prevalence[["age_lower", "age_upper", "time_lower", "time_upper", "sex_id", "node_id", "density", "weight"]]
+    emr["measure"] = "mtexcess"
+
+    def emr_mean(prevalence_measurement):
+        if (
+            prevalence_measurement.time_lower == prevalence_measurement.time_upper
```
adolgert: Doing the equivalence test multiple times might be considered premature pessimization. What would you think of assigning a variable that says time_is_point_estimate or age_is_point_estimate? That way we could reuse those in the elif below.

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+    csmr = _prepare_csmr(execution_context, csmr)
+
+    emr = _calculate_emr_from_csmr_and_prevalence(csmr, prevalence)
+
+    model_context.input_data.observations = pd.concat([model_context.input_data.observations, emr])
+
+
+def _calculate_emr_from_csmr_and_prevalence(all_csmr, all_prevalence):
+    MATHLOG.debug(
+        dedent(
```
adolgert: I love that you know dedent. I don't use this enough.

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+    """
+        )
+    )
+    excess_mortality_chunks = []
+    non_interpolatable_groups = all_prevalence[["sex_id", "node_id"]].drop_duplicates()
+    for (sex_id, node_id) in non_interpolatable_groups.values:
+        csmr = all_csmr.query("sex_id == @sex_id and node_id == @node_id")
+        prevalence = all_prevalence.query("sex_id == @sex_id and node_id == @node_id")
+        if csmr.empty:
+            MATHLOG.debug(
```
adolgert: Odd question. Running `python -O` turns off debug statements, right? No, it turns off `__DEBUG__` but leaves logging debug statements. nvmd.

src/cascade/stats/__init__.py
```diff
@@ -0,0 +1,40 @@
+from cascade.stats.compartmental import (
```
adolgert: Nice to pull these up to the top.

tests/input_data/test_emr.py
```diff
@@ -0,0 +1,194 @@
...
+    csmr = pd.DataFrame(csmr_rows)
+
+    emr = _emr_from_sex_and_node_specific_csmr_and_prevalence(csmr, SPAN_PREVALENCE)
+
+    assert len(emr) == len(SPAN_PREVALENCE)
+
+    for (_, pr), (_, er) in zip(SPAN_PREVALENCE.iterrows(), emr.iterrows()):
+        pmean = pr["mean"]
+        cmean = csmr_mean((pr["age_lower"] + pr["age_upper"]) / 2, (pr["time_lower"] + pr["time_upper"]) / 2)
+        assert er["mean"] - cmean / pmean < er["standard_error"]
```
adolgert: Does this pass now?

alec-deason:
I added tests for my interpolation logic in tests/input_data/test_emr.py:test_interpolators__across_ages Can someone glance over those to make sure they aren't crazy. I'm trying to catch gross errors but I don't want to make them too loose.

src/cascade/input_data/emr.py
alec-deason: I don't think I would ask you to do that, @adolgert might? I think the exact value here isn't important and is unlikely to change or be tuned.

src/cascade/input_data/emr.py
alec-deason: Complain at Drew. He wrote it. It should render real nice in the Sphinx docs.

src/cascade/input_data/emr.py
alec-deason: Yep, this deserves a docstring.

src/cascade/input_data/db/demographics.py
alec-deason: Ok, that makes sense to me.

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+
+from scipy.interpolate import InterpolatedUnivariateSpline, LSQBivariateSpline
+
+import numpy as np
+import pandas as pd
+
+from cascade.input_data.db.mortality import get_cause_specific_mortality_data
+from cascade.input_data.db.demographics import age_groups_to_ranges, get_mean_years
+from cascade.stats import meas_bounds_to_stdev
+
```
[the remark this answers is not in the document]
alec-deason: It's just that the logger import and use is a common block that's repeated over and over. It is it's own lexical item. If python had macros I'd make this a macro

src/cascade/input_data/emr.py
alec-deason: Sure

src/cascade/input_data/emr.py
alec-deason: Learned it from Mike Richards. It's great.

src/cascade/input_data/emr.py
alec-deason: Yeah, it would be odd for it to effect the logger.

tests/input_data/test_emr.py
alec-deason: Don't ask me, ask Travis. Which is to say, yes. That is the right way to loosen the test, right?

PR #93, opened 2018-11-13
Policies

alec-deason:
This introduces a simple system for configuring policies using our existing form parsing tools. It also adds one policy for EMR calculation.


src/cascade/input_data/db/demographics.py
```diff
@@ -30,20 +31,38 @@ def get_locations(execution_context):
...
     groups = get_age_groups(execution_context)
-    with_group = data.merge(groups, on="age_group_id").drop(columns="age_group_id")
+    with_group = data.merge(groups, on="age_group_id")
+    if not keep_age_group_id:
+        with_group = with_group.drop(columns="age_group_id")
     return with_group.rename(columns={"age_group_years_start": "age_lower", "age_group_years_end": "age_upper"})
 
 
-def get_all_age_spans(execution_context):
+def get_mean_years(execution_context):
```
nrhawkins: This doesn't really bother me, but I feel a little question arising (before reading the docstring), does mean_years mean something about a span like 2005-2010, though years plural, sounds like it doesn't, or does it mean something about an age group like 5-10?  =7.5 years

alec-deason: Neither. It is the average age that you expect someone within this age group to have minus the starting age of the group. If you can think of a better name for that, I'm all ears because I don't have a good one.

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+def _prepare_prevalence(observations):
+    prevalence = observations.query("measure == 'prevalence' and sex_id in [1, 2, 3]")
+
+    zero_means = prevalence["mean"] == 0
+    if np.any(zero_means):
+        not_quite_zero = 1e-6
+        MATHLOG.debug(f"{np.sum(zero_means)} prevalence rows with 0 means. Changing them to {not_quite_zero}.")
+        prevalence.loc[zero_means, "mean"] = not_quite_zero
+
+    MATHLOG.debug(f"{len(prevalence)} prevalence observations avaliable, exluding sex_id 4")
```
nrhawkins: avaliable -> available

exluding -> excluding

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+    mean["both"] = LSQBivariateSpline(
+        csmr.age, csmr.time, csmr["mean"], sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
+    )
+    stderr["both"] = LSQBivariateSpline(
+        csmr.age, csmr.time, csmr.standard_error, sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
+    )
+
+    # csmr_by_age = csmr.sort_values("age").groupby("age").mean().reset_index()
+    # mean["age"] = InterpolatedUnivariateSpline(csmr_by_age.age, csmr_by_age["mean"], k=1)
+    # stderr["age"] = InterpolatedUnivariateSpline(csmr_by_age.age, csmr_by_age.standard_error, k=1)
```
nrhawkins: I think a certain someone would ask me to remove these lines, if I left them in?

src/cascade/input_data/emr.py
```diff
@@ -0,0 +1,219 @@
...
+            and prevalence_measurement.age_lower == prevalence_measurement.age_upper
+        ):
+            # No integral, just interpolate to a point
+            csmr_mean = mean_interp["both"](prevalence_measurement.age_lower, prevalence_measurement.time_lower)
+            csmr_stderr = stderr_interp["both"](prevalence_measurement.age_lower, prevalence_measurement.time_lower)
+        elif prevalence_measurement.time_lower == prevalence_measurement.time_upper:
+            # Integrate over age
+            divisor = prevalence_measurement.age_upper - prevalence_measurement.age_lower
+
+            csmr_mean = mean_interp["age"].integral(prevalence_measurement.age_lower, prevalence_measurement.age_upper)
```
nrhawkins: I think the "age" case is commented out above in "def _make_interpolators(csmr):"?

src/cascade/input_data/emr.py
alec-deason: Well, if it were me who were that certain someone, I would ask you to test your code better because that's not just a bug but a stupid one.

src/cascade/input_data/emr.py
alec-deason: Yep

tests/model/test_compartmental.py
```diff
@@ -1,7 +1,9 @@
 import numpy as np
+
 from scipy.stats import gamma
 
-from cascade.model import (
+from cascade.stats import (
+    DemographicInterval,
```
nrhawkins: above in stats/__init__.py we have:
from cascade.stats.demography import DemographicInterval

Do we have that, so that here we don't have to have that?  Also, same goes for test_demography.py?

src/cascade/input_data/db/demographics.py
adolgert: Alec, did you get rid of the age spans call? I'm sure we will need it to know the start and finish of each span. Maybe, if and when it's surely necessary, then it's easy to put in.

src/cascade/input_data/configuration/form.py
```diff
@@ -189,6 +189,7 @@ class Model(Form):
         default="iota_pos_rho_zero",
         display="(Advanced) Rate case",
     )
+    use_weighted_age_group_midpoints = OptionField([1, 0], default=1, constructor=int, nullable=True)
```
adolgert: OK, so this is an example of use.

src/cascade/input_data/configuration/builder.py
```diff
@@ -56,6 +57,8 @@ def initial_context_from_epiviz(configuration):
     context.parameters.rate_case = configuration.model.rate_case
     context.parameters.minimum_meas_cv = configuration.model.minimum_meas_cv
 
+    context.policies = policies_from_settings(configuration)
```
adolgert: Ah! I was concerned that you need settings to have policies. That would mean that you need the IHME part in order to run the non-IHME code, which would be bad. This step here is enough to separate the two.

tests/model/test_compartmental.py
alec-deason: Yeah, there is a convention in python that a sub-package pulls important names from it's modules into the sub-packages root level for easy importing and to make the public API clear. I tend not to do that, more from laziness than rational choice, so a lot of our code isn't written that way. It's happening in cascade.stats because of Drew.

alec-deason:
Natalie, most of your comments were actually on code that come in on the EMR branch because I did this in an unorganized way. I've pushed my changes there too.

src/cascade/input_data/db/demographics.py
alec-deason: Look down, the diff is just confused.

src/cascade/input_data/configuration/form.py
alec-deason: Oh, no. I should have moved that. It's in the wrong spot.

PR #94, opened 2018-11-14
Get the right CSMR data for a given Cod version

alec-deason:
Peers into the GBD databases to get the magic version number for CSMR

### DIFF
--- a/src/cascade/input_data/db/csmr.py
+++ b/src/cascade/input_data/db/csmr.py
@@ -3,7 +3,7 @@
 import pandas as pd
 
 from cascade.core.db import cursor, db_queries
-from cascade.input_data.db import GBD_ROUND_ID, METRIC_IDS, MEASURE_IDS
+from cascade.input_data.db import GBD_ROUND_ID, METRIC_IDS, MEASURE_IDS, GBDDataError
 
 
 from cascade.core.log import getLoggers
@@ -31,6 +31,12 @@ def _csmr_in_t3(execution_context):
 
 
 def _gbd_process_version_id_from_cod_version(cod_version):
+    """Central comp uses process_version_id to track data versions independently
+    from the versioning systems of the tools that produced the data. Thus we
+    need to map from CODcorrect's versions to a process_version_id to do
+    version constrained lookups using central comp's tools.
+    """
+
     query = """
     SELECT gbd.gbd_process_version_metadata.gbd_process_version_id from gbd.gbd_process_version_metadata
         JOIN  gbd.gbd_process_version ON gbd.gbd_process_version_metadata.gbd_process_version_id =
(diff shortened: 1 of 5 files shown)


src/cascade/input_data/db/csmr.py
```diff
@@ -29,13 +30,36 @@ def _csmr_in_t3(execution_context):
...
+
+
 def _get_csmr_data(execution_context):
 
     cause_id = execution_context.parameters.add_csmr_cause
     parent_loc = execution_context.parameters.location_id
 
     keep_cols = ["year_id", "location_id", "sex_id", "age_group_id", "val", "lower", "upper"]
 
+    process_version_id = _gbd_process_version_id_from_cod_version(execution_context.parameters.cod_version)
```
nrhawkins: Maybe we should add some comments here to say a little something about what process_version_id is?

src/cascade/input_data/db/csmr.py
```diff
@@ -92,15 +116,11 @@ def load_csmr_to_t3(execution_context) -> bool:
 
     if _csmr_in_t3(execution_context):
         CODELOG.info(
-            f"csmr data for model_version_id {model_version_id} "
-            f"on '{database}' already exists, doing nothing."
+            f"csmr data for model_version_id {model_version_id} " f"on '{database}' already exists, doing nothing."
```
nrhawkins: If these are being changed to fit on one line, should the f""'s be combined into one?  Just wondering.

src/cascade/input_data/db/csmr.py
```diff
@@ -29,13 +30,36 @@ def _csmr_in_t3(execution_context):
     return exists == 1
 
 
+def _gbd_process_version_id_from_cod_version(cod_version):
+    query = """
+    SELECT gbd.gbd_process_version_metadata.gbd_process_version_id from gbd.gbd_process_version_metadata
+        JOIN  gbd.gbd_process_version ON gbd.gbd_process_version_metadata.gbd_process_version_id =
+                                         gbd.gbd_process_version.gbd_process_version_id
+        WHERE gbd_process_id = 3                     -- codcorrect process
```
adolgert: Comments inside the SQL!

src/cascade/input_data/db/csmr.py
```diff
@@ -29,13 +30,36 @@ def _csmr_in_t3(execution_context):
...
+              and val = %(cod_version)s
+              and gbd_process_version_status_id = 1  -- marked best
+    """
+
+    with cursor(database="gbd") as c:
+        c.execute(query, args={"cod_version": cod_version})
+        result = c.fetchone()
+
+    if result is None:
+        raise ValueError(f"No best gbd_process_version_id for cod version {cod_version}")
```
adolgert: If there weren't a best process version, this will make it fail... I don't know that it should recover. There is an InputDataError(Exception) for things the modeler can fix. Does this fall in that category? If not, it should have a category of things central comp wants to know.

alec-deason: I new in advance that you would complain about this but even so I didn't do it right. Such is foresight. How about a GBDDataError?

src/cascade/input_data/db/csmr.py
adolgert: I like it actually.

src/cascade/input_data/db/csmr.py
adolgert: That sounds great.

PR #99, opened 2018-11-27
Bugfix/estimate emr fails

adolgert:
A bug got into the tests that use --ihme from the new policies. This makes the test run with the policies. It also found a hidden bug where emr would fail if there were no prevalence measurements, so it fixes that.


tests/input_data/configuration/test_build_mortality.py
```diff
@@ -9,7 +9,9 @@
 @pytest.mark.parametrize("sexes", [1, 2])
 def test_add_mortality_data(ihme, sexes):
     ec = make_execution_context(model_version_id=265976, gbd_round_id=5)
+    ec.policies = dict(use_weighted_age_group_midpoints=0)
```
alec-deason: Ugg. This is awful. I never meant for this  to be an attribute on ExecutionContext. I'll make a PR to clean this up.

PR #103, opened 2018-12-03
Feature/dismodat 370 country covariates assignment and notebook

nrhawkins:
This PR contains the code for the country covariates assignment to measurements, which uses the griddata interpolator.  I removed the drafts of the corresponding notebooks, so that this part can be reviewed while one of those gets reworked.


src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+    0.667 in age_interval (no), 95 in age_interval (no), 50.1 in age_interval (yes)
+
+    Returns:
+        Interval: of ages in the covariates data frame
+    """
+    age_interval = it.empty()
+
+    age_groups = set(zip(covariates["age_lower"], covariates["age_upper"]))
+
+    for age in age_groups:
```
alec-deason: Instead of forcing the dataframe into a set above I would use `.drop_duplicates` there and `.iterrows` here. For small dataframes they're equivalent but for larger ones staying in pandas makes fewer copies which can be meaningful.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+def get_measurement_data_by_sex(measurements):
+    """Split the measurement data by sex values found in the measurements.
+    Args:
+        measurements (pandas.DataFrame): data for a specific measurement
+    Returns:
+        dict: possible sex keys (-0.5, 0, 0.5) and measurement data as values
+    """
+    measurements_by_sex = {}
+
+    for sex in (FEMALE, MALE, BOTH):
```
alec-deason: NEITHER?

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+        measurements, covariates, covar_at_dims, covar_age_interval):
+    """
+    Use the measurements data by sex as input to the corresponding interpolator
+    to assign a covariate value to the measurement.
+
+    Returns:
+        pd.Series: One row for every row in the measurements.
+    """
+
+    covariates["avg_age"] = covariates[["age_lower", "age_upper"]].mean(axis=1)
```
alec-deason: Should we be using mortality weighted midpoints here?

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+    covariates["avg_time"] = covariates[["time_lower", "time_upper"]].mean(axis=1)
+
+    covariates_by_sex = get_covariate_data_by_sex(covariates)
+
+    measurements["avg_age"] = measurements[["age_lower", "age_upper"]].mean(axis=1)
```
alec-deason: Same question as above. There's some code for this in EMR: https://github.com/ihmeuw/cascade/blob/develop/src/cascade/input_data/emr.py#L40

Also this for the relevant policy: `model_context.policies["use_weighted_age_group_midpoints"]`

We should probably extract this logic out into a shared utility since this as well as EMR need it and probably other places too.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -1,10 +1,15 @@
...
 """
+
 from collections.__init__ import defaultdict
 
+import numpy as np
 import pandas as pd
 from scipy import spatial
+from scipy.interpolate import griddata
+
+import intervals as it
```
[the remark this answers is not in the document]
alec-deason: It appears that this isn't even used. You make an Interval object but never actually do anything with it?

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+    """
+    Determines if the input covariate data is by_age and/or by_time.
+
+    Returns:
+        dictionary: keys indicate if age or time are points (0d) vs. 1d
+    """
+
+    covar_at_dims = {}
+
+    covar_at_dims["age_1d"] = len(set(zip(covariates["age_lower"], covariates["age_upper"]))) > 1
```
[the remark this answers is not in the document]
nrhawkins: ok, did this!

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -217,3 +226,187 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
...
+
+def compute_covariate_age_time_dimensions(covariates):
+    """
+    Determines if the input covariate data is by_age and/or by_time.
+
+    Returns:
+        dictionary: keys indicate if age or time are points (0d) vs. 1d
+    """
+
+    covar_at_dims = {}
```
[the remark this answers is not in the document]
nrhawkins: ok, I used namedtuple, thanks!

src/cascade/input_data/configuration/construct_country.py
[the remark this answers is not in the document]
nrhawkins: We discussed at stand-up, keeping this dependency, at least, until it gives us problems.

The interval is used in:  assign_interpolated_covariate_values

src/cascade/input_data/configuration/construct_country.py
nrhawkins: ok, did this!

src/cascade/input_data/configuration/construct_country.py
nrhawkins: I added NEITHER, and treated it like BOTH.

src/cascade/input_data/configuration/construct_country.py
nrhawkins: Not doing this, at least not yet.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -1,19 +1,30 @@
...
 from cascade.input_data.db.demographics import get_all_age_spans
 from cascade.input_data.configuration.construct_study import CovariateRecords
-from cascade.core import getLoggers
 
 CODELOG, MATHLOG = getLoggers(__name__)
 
+FEMALE = -0.5
+MALE = 0.5
+BOTH = 0
+NEITHER = 0.1
```
alec-deason: I know I called out NEITHER and we do need to support it (maybe? I'm still not sure what exactly the right thing is) but since we haven't agreed on how to represent it in the databases let's actually skip it for now. Sorry to waffle. At very least it can't be 0.1, we'll have to readjust how the whole system works to handle it I think.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -23,7 +23,6 @@
 FEMALE = -0.5
 MALE = 0.5
 BOTH = 0
-NEITHER = 0.1
```
alec-deason: Thanks. Sorry for the confusion

PR #104, opened 2018-12-04
infer rate cases and validate fixed effects lower bounds

alec-deason:
The "rate case" configuration in the "Advanced" tab is confusing and redundant. This change adds code to infer the correct rate case to use based on the priors the user supplied. It also adds stricter validation on the bounds for fixed effect priors.


alec-deason:
I actually made an incorrect assumption in the validation part of this PR which I need to fix. I'll try to get that done this afternoon.

PR #108, opened 2018-12-07
Mortality data and smoothings for children

alec-deason:
This gets us mortality data for the child nodes (not all descendants) and actual support for the nslist and nspairs system which we punted on before.

### DIFF
--- a/src/cascade/dismod/serialize.py
+++ b/src/cascade/dismod/serialize.py
@@ -15,6 +15,7 @@ from cascade.model.grids import unique_floats
 from cascade.input_data.db.locations import get_location_hierarchy_from_gbd
 
 from cascade.core.log import getLoggers
+
 CODELOG, MATHLOG = getLoggers(__name__)
 
 
@@ -95,13 +96,13 @@ def model_to_dismod_file(model, execution_context):
     # cases by ID and save. Any covariates have to exist at this point.
     bundle_fit.avgint = make_avgint_table(model, integrand_id_func, location_to_node_func, covariate_renames)
 
-    bundle_fit.rate, rate_id_func = make_rate_table(model, smooth_id_func)
-
-    bundle_fit.mulcov = make_mulcov_table(
-        model, smooth_id_func, rate_id_func, integrand_to_id, cov_col_id_func
+    bundle_fit.rate, rate_id_func, bundle_fit.nslist, bundle_fit.nslist_pair = make_rate_and_nslist_tables(
+        model, smooth_id_func, location_to_node_func
     )
 
-    bundle_fit.option = make_option_table(model)
+    bundle_fit.mulcov = make_mulcov_table(model, smooth_id_func, rate_id_func, integrand_to_id, cov_col_id_func)
+
+    bundle_fit.option = make_option_table(model, location_to_node_func)
 
     return bundle_fit
 
@@ -187,7 +188,8 @@ def make_data_table(model_context, node_table, covariate_renames):
                 "nu",
                 "hold_out",
                 "data_name",
-            ] + list(covariate_renames.values())
+            ]
+            + list(covariate_renames.values())
         )
     return total_data
 
@@ -203,10 +205,9 @@ def observations_to_data(observations_df, node_table):
     """Turn an internal format into a Dismod format."""
     # Don't make the data_name here because could convert multiple observations.
     observations_df = observations_df.reset_index()
-    observations_df["node_id"] = observations_df.merge(node_table,
-                                                       left_on="node_id",
-                                                       right_on=node_table.c_location_id
-                                                       ).node_id_y
+    observations_df["node_id"] = observations_df.merge(
+        node_table, left_on="node_id", right_on=node_table.c_location_id
+    ).node_id_y
     transformed = observations_df.assign(
         integrand_id=observations_df["measure"].apply(lambda x: IntegrandEnum[x].value),
         density_id=observations_df["density"].apply(lambda x: x.value),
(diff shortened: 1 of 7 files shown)


src/cascade/dismod/serialize.py
```diff
@@ -506,28 +527,52 @@ def smooth_id_func(smooth):
...
+    nslist_rows = []
+    nspairs_rows = []
     for rate_id, rate in enumerate(context.rates):
-        if len(rate.child_smoothings) > 1:
-            raise NotImplementedError("Multiple child smoothings not supported yet")
-
-        rows.append(
+        nslist_id = np.NaN
+        child_smooth_id = np.NaN
+        if rate.child_smoothings:
```
adolgert: What happens if the rate doesn't have any child smoothings? Every rate needs some smoothing. Does that mean it shares the parent smoothing, and should I see some registration of that parent smoothing as being associated with that rate? I'm concerned this might miss a case.

src/cascade/executor/epiviz_runner.py
```diff
@@ -99,8 +101,23 @@ def add_omega_constraint(model_context, execution_context, sex_id):
...
+    MATHLOG.debug(f"Add {parent_asdr.shape[0]} omega constraints from age-standardized death rate data to the parent.")
+
+    location_hierarchy = get_location_hierarchy_from_gbd(execution_context)
+    location = location_hierarchy.get_node_by_id(execution_context.parameters.location_id)
+    children = {d.id for d in location.children}  # noqa: F841
+    children_asdr = asdr.query("node_id in @children")
+    model_context.rates.omega.child_smoothings = [
+        (node_id, build_constraint(child_asdr))
+        for node_id, child_asdr in children_asdr.groupby('node_id')
+    ]
```
adolgert: This is succinct for calling build_constraint many times.

src/cascade/input_data/db/asdr.py
```diff
@@ -32,14 +33,16 @@ def _asdr_in_t3(execution_context):
...
 
-    parent_loc = execution_context.parameters.location_id
-
     demo_dict = db_queries.get_demographics(gbd_team="epi", gbd_round_id=execution_context.parameters.gbd_round_id)
     age_group_ids = demo_dict["age_group_id"]
     sex_ids = demo_dict["sex_id"]
 
+    location_hierarchy = get_location_hierarchy_from_gbd(execution_context)
+    location = location_hierarchy.get_node_by_id(execution_context.parameters.location_id)
+    location_and_children = [d.id for d in location.children] + [location.id]
```
adolgert: This is the second use of the same pattern to get children.

src/cascade/input_data/db/csmr.py
```diff
@@ -60,16 +61,19 @@ def _gbd_process_version_id_from_cod_version(cod_version):
...
     cause_id = execution_context.parameters.add_csmr_cause
-    parent_loc = execution_context.parameters.location_id
 
     keep_cols = ["year_id", "location_id", "sex_id", "age_group_id", "val", "lower", "upper"]
 
     process_version_id = _gbd_process_version_id_from_cod_version(execution_context.parameters.cod_version)
 
+    location_hierarchy = get_location_hierarchy_from_gbd(execution_context)
+    location = location_hierarchy.get_node_by_id(execution_context.parameters.location_id)
+    location_and_children = [d.id for d in location.children] + [location.id]
```
adolgert: And third example of same pattern.

src/cascade/dismod/serialize.py
```diff
@@ -217,9 +218,22 @@ def observations_to_data(observations_df, node_table):
...
+        "hold_out",
+        "meas_value",
+        "meas_std",
+        "eta",
+        "nu",
+        "age_lower",
+        "age_upper",
+        "time_lower",
+        "time_upper",
+    }
```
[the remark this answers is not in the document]
alec-deason: I think this (and a few other style changes) are because I ran the pretty printer on this file. I've been trying to be careful to only do that in cases where it won't clutter up PRs or interfere with other folks' style choices but I seem to have run it here. That said I think I'm leaning towards long style as well. I used to argue against it but it looks ok to me now.

src/cascade/dismod/serialize.py
```diff
@@ -506,28 +527,52 @@ def smooth_id_func(smooth):
     return smooth_table, grid_table, smooth_id_func
 
 
-def make_rate_table(context, smooth_id_func):
+def make_rate_and_nslist_tables(context, smooth_id_func, location_to_node_func):
```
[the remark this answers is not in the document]
alec-deason: Ok, I'll comment. The short answer is because if I do them together I can do it in one fairly simple pass but if I separate them I need more passes and more complexity (I think).

src/cascade/dismod/serialize.py
alec-deason: From Brad's docs: "If child_smooth_id is not null, child_nslist_list must be null. If both are null, there are no random effects . for this rate."

It seems like if the modeler supplies no random effects smoothings, which is the else case here, they should get the "no random effects for this rate" case? Or am I misunderstanding something?

src/cascade/executor/epiviz_runner.py
alec-deason: Please note that I didn't do it all on one line! I'm growing.

src/cascade/input_data/db/csmr.py
alec-deason: Yes, you are right. I was avoiding thinking about how to clean that up but it's silly to have the repeated code.

src/cascade/dismod/serialize.py
```diff
@@ -506,28 +527,52 @@ def smooth_id_func(smooth):
...
     for rate_id, rate in enumerate(context.rates):
-        if len(rate.child_smoothings) > 1:
-            raise NotImplementedError("Multiple child smoothings not supported yet")
-
-        rows.append(
+        nslist_id = np.NaN
+        child_smooth_id = np.NaN
+        if rate.child_smoothings:
+            if len(rate.child_smoothings) == 1 and rate.child_smoothings[0][0] is None:
+                # This is a blanket smoothing for all children
```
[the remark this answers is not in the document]
alec-deason: Now that depends on how exactly you define smooth.

PR #118, opened 2018-12-17
fixed bug in var iterator and created test for it

adolgert:
Tiny, tiny PR. Tiny PR I love you. Tiny PR love me too. Tiny, tiny PR.

The var grid iterator should have iterated over random fields. There can be more than one random effect using the same smooth grid id. This fixes that and adds a test that exercises that possibility.

### DIFF
--- a/src/cascade/model/operations.py
+++ b/src/cascade/model/operations.py
@@ -8,8 +8,8 @@ from numpy import nan
 import pandas as pd
 
 
-def _var_grid_iterator(var_df):
-    r"""Iterate over grids in the var table, returning the type of grid and the
+def random_field_iterator(var_df):
+    r"""Iterate over random fields in the var table, returning the type of grid and the
     sub-section of the var table corresponding to that grid.
 
      * There is an MRF for every (``rate_id``, ``node_id``) combination,
@@ -21,16 +21,6 @@ def _var_grid_iterator(var_df):
        applied to the measured value, measured stdev, or the rate. These are the
        covariates :math:`(\alpha, \beta, \gamma)`.
 
-    This function validates that we have an exhaustive list of what is in
-    a ``smooth_grid``.
-
-    .. note::
-
-        This walks the var table, but it could find all smooth grids by
-        looking at the rate table and mulcov table. Are all smooth grids
-        in the var table? It looks like even rates that are constrained
-        to be constant are in the var table.
-
     Args:
         var_df (pd.DataFrame): The DB.var table from the Dismod-AT database.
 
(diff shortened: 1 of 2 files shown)


src/cascade/model/operations.py
```diff
@@ -21,68 +21,33 @@ def _var_grid_iterator(var_df):
...
+        if var_type == "mulcov_meas_value":
+            kind = "meas_value"
+        elif var_type == "mulcov_meas_std":
+            kind = "meas_std"
         else:
-            raise RuntimeError(f"Grid has unknown type for smooth_id={smooth_id}")
+            raise RuntimeError(f"Unknown var_type {var_type} in var table")
+        yield kind, (smooth_id, integrand_id, covariate_id), sub_grid_df
+
+    for alpha_index, sub_grid_df in var_df.groupby(["smooth_id", "rate_id", "covariate_id"]):
```
alec-deason: Either I misunderstand what this is doing or I misunderstand how groupby works. Why do these groupbys not return overlapping results?

src/cascade/model/operations.py
```diff
@@ -21,68 +21,33 @@ def _var_grid_iterator(var_df):
...
-            yield "meas_std", (smooth_id, integrand_id, covariate_id), sub_grid_df
-
-        elif not rate_value.empty:
-            rate_id, covariate_id = [int(rval_id) for rval_id in
-                                     rate_value.iloc[0]]
-            yield "rate_value", (smooth_id, rate_id, covariate_id), sub_grid_df
-
+    # groupby ignores every record where any of the keys are Null, so it
+    # only gets the rates.
+    for rate_index, sub_grid_df in var_df.groupby(["smooth_id", "rate_id", "node_id"]):
```
nrhawkins: Does this get rates and rate_value?

tests/model/test_operations.py
```diff
@@ -135,6 +136,33 @@ def draws_at_value(mean, std, cnt):
...
+def test_var_grid_iterator():
+    # parent rate 0, parent rate 1,
+    # child 1 rate 0, child 2 rate 0, with shared smooth grid
+    # child 1 rate 1, different smooth grid.
+    # Note there can be two rates with the _same smooth id_ but different nodes.
+    var_df = pd.DataFrame(dict(
+        var_id=[0, 1, 2, 3, 4, 5, 6, 7],
+        smooth_id=[0, 1, 2, 2, 3, 4, 5, 6],
+        var_type=["rate", "rate", "rate", "rate", "rate", "mulcov_meas_value",
+                  "mulcov_rate_value", "mulcov_meas_std"],
```
nrhawkins: Should rate_value be in here too?

src/cascade/model/operations.py
adolgert: Groupby ignores rows with Nulls in the given columns. That's why. I put a comment to that effect above.

src/cascade/model/operations.py
adolgert: It doesn't get rate value because the node_id is null for rate value, so it doesn't show up in the gruop by.

tests/model/test_operations.py
adolgert: I think "mulcov_rate_value" is what you call rate value. Meaning, the db uses "mulcov_rate_value" but we call them "rate_value" in another part of the DB.

PR #119, opened 2018-12-18
Bugfix/omega constraint as effect

adolgert:
Converts child omega constraint into a random effect by changing its mean value to the log(child mean / parent mean) for all children.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -91,10 +91,21 @@ def add_mortality_data(model_context, execution_context, sex_id):
 
 
 def add_omega_constraint(model_context, execution_context, sex_id):
-    """
+    r"""
     Adds a constraint to other-cause mortality rate. Removes mtother,
     mtall, and mtspecific from observation data. Uses
     :py:func:`cascade.input_data.configuration.builder.build_constraint` to make smoothing priors.
+
+    Child constraints are constrained random effects, so they are offsets
+    from the parent omega constraint, as
+
+    .. math::
+
+        \omega_c(a,t) = \omega_p(a,t) e^{u_c(a,t)}
+
+    It is the :math:`u_c` that we put into the child smoothings. We define
+    these grids so that all grid points match. At each grid point,
+    :math:`u_c=\log (c/p)` with c for child, p for parent.
     """
     MATHLOG.debug(f"Add omega constraint from age-standardized death rate data.")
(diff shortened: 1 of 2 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -91,15 +92,28 @@ def add_mortality_data(model_context, execution_context, sex_id):
...
     """
-    MATHLOG.debug(f"Add omega constraint from age-standardized death rate data.")
-    MATHLOG.debug("Assigning standard error from measured upper and lower.")
+    MATHLOG.debug(f"Add omega constraint from age-standardized death rate data. "
+                  f"Assigning standard error from measured upper and lower.")
+    # Call this the long way so it can be monkeypatched in testing.
     asdr = meas_bounds_to_stdev(
-        age_groups_to_ranges(execution_context, get_age_standardized_death_rate_data(execution_context))
+        age_groups_to_ranges(execution_context,
+                             cascade.input_data.db.mortality.get_age_standardized_death_rate_data(execution_context))
```
alec-deason: Why fully qualified?

src/cascade/executor/epiviz_runner.py
```diff
@@ -116,9 +130,17 @@ def add_omega_constraint(model_context, execution_context, sex_id):
 
     children = get_descendents(execution_context, children_only=True)  # noqa: F841
     children_asdr = asdr.query("node_id in @children")
+    # Transform the children to be the random effect for the rate.
+    parent_value = parent_asdr[["age_lower", "time_lower", "mean"]].rename({"mean": "parent_mean"}, axis=1)
+    parent_and_child = children_asdr.merge(parent_value, how="left", on=["age_lower", "time_lower"])
+    # This could result in an Inf value, but that's a legal value, so...
+    children_effects = parent_and_child.assign(
```
alec-deason: Oh, nice. This is less complicated than I'd feared.

src/cascade/executor/epiviz_runner.py
adolgert: The pytest mocking fixture changes the getattr on the import. It doesn't work if the function is imported directly. Annoying. I put a comment in to that effect.

src/cascade/executor/epiviz_runner.py
adolgert: It gets simple when parent and effects are on the same grid, and we defined them, which puts them on the same grid.

PR #120, opened 2018-12-18
Feature/bundle data from disk

alec-deason:
This is not super complicated but makes a bunch of small changes to support loading bundles off disk and skipping the tier3 cache in development situations.


src/cascade/core/db.py
```diff
@@ -128,3 +130,14 @@ def latest_model_version(execution_context):
         else:
             raise RuntimeError(
                 f"No model version for modelable entity id {model_id} in database.")
+
+
+def dataframe_from_disk(path):
+    """ Load the file at `path` as a pandas dataframe.
+    """
+    if any(path.endswith(extension) for extension in [".hdf", ".h5", ".hdf5", ".he5"]):
```
adolgert: I've never seen he5. Where have you seen it?

src/cascade/executor/epiviz_runner.py
```diff
@@ -72,8 +76,13 @@ def add_mortality_data(model_context, execution_context, sex_id):
     MATHLOG.debug(f"Creating a set of mtspecific observations from IHME CSMR database.")
     MATHLOG.debug("Assigning standard error from measured upper and lower.")
 
+    if execution_context.parameters.tier == 3:
+        raw_csmr = normalize_mortality_data(get_cause_specific_mortality_data(execution_context))
```
adolgert: get_csmr and get_cause_specific_mortality are the same meaning. Maybe one should be get_frozen and get_nonfrozen, as with asdr below.

src/cascade/executor/epiviz_runner.py
```diff
@@ -388,12 +415,19 @@ def main(args):
...
     """
     ec = make_execution_context()
+
     settings = load_settings(ec, args.meid, args.mvid, args.settings_file)
 
     if settings.model.drill != "drill":
         raise NotImplementedError("Only 'drill' mode is currently supported")
 
     add_settings_to_execution_context(ec, settings)
+    if args.skip_cache:
```
adolgert: This is an interesting option.

src/cascade/input_data/configuration/construct_study.py
```diff
@@ -120,7 +121,10 @@ def add_special_study_covariates(covariate_records, model_context, sex_id):
 
 def get_bundle_study_covariates(model_context, bundle_id, execution_context, tier):
     # Sparse data is specified as a list of seq IDs that have a particular study covariate.
-    sparse_covariate_data = get_study_covariates(execution_context, bundle_id, tier=tier)
+    if execution_context.parameters.bundle_study_covariates_file:
+        sparse_covariate_data = dataframe_from_disk(execution_context.parameters.bundle_study_covariates_file)
```
adolgert: Right. Study covariates too.

src/cascade/input_data/db/asdr.py
```diff
@@ -97,6 +87,13 @@ def _upload_asdr_data_to_tier_3(cursor, model_version_id, asdr_data):
         )
     """
 
+    age_group_data = db_queries.get_age_metadata(
+        age_group_set_id=AGE_GROUP_SET_ID, gbd_round_id=execution_context.parameters.gbd_round_id
+    )[["age_group_id", "age_group_years_start", "age_group_years_end"]]
+
+    age_group_data.columns = ["age_group_id", "age_lower", "age_upper"]
```
adolgert: If the order of the columns doesn't match this, does it work OK? I'd feel more comfortable using `rename({}, axis=1)`

src/cascade/core/db.py
alec-deason: List of extensions used by hdf on wikipedia. I've not seen it either. We can take it out if you want to be tighter.

src/cascade/executor/epiviz_runner.py
alec-deason: It's actually important because if you do a lot of testing using manipulated json files it's easy to get the tier3 caches into a state that is inconsistent with the models you are trying to run. And then failure.

src/cascade/input_data/db/asdr.py
alec-deason: It is guaranteed to match because of the line above but I agree, rename is better.

src/cascade/executor/epiviz_runner.py
alec-deason: Yeah

src/cascade/executor/epiviz_runner.py
adolgert: Makes sense.

alec-deason:
Ok. I stuck in some description of what should go in the bundles. It's not super in depth but it's there.

PR #124, opened 2018-12-24
fixed sqlalchemy pool timeout and installation permissions problems

adolgert:
1. Our database connections were sometimes running out of resources. This fixes the resource problem by adding a single line that closes connections in a central part of our library. The problem showed up during acceptance testing but has not been seen in production, to my knowledge.
2. Installation could fail silently in two ways: a user would be unable to run, or another team member would be unable to complete an install. This PR modifies the installation script to ensure permissions are appropriate for the team and to ensure that it creates a fully-compiled environment for the user.
3. Installation required passing unit tests for an install. This now adds a requirement that acceptance tests run in order to complete an install.

Not a lot of code, but it fixes some things.

### DIFF
--- a/scripts/automation/install.sh
+++ b/scripts/automation/install.sh
@@ -56,7 +56,7 @@ fi
 # Download the Cascade project from github
 
 CASCADE_HOME=/ihme/code/dismod_at
-CASCADE_DEVELOP_DIR=${CASCADE_HOME}/cascade_develop
+CASCADE_DEVELOP_DIR=${CASCADE_HOME}/clone_for_install
 
 GITHUB="https://github.com/ihmeuw/cascade.git"
 
@@ -95,8 +95,10 @@ if [ "$?" -ne "0" ]; then
     exit 7
 fi
 
-# Install the virtual environment to be the "current" if the tests pass
-(cd "${CASCADE_DEVELOP_DIR}/tests" && pytest)
+# Install the virtual environment to be the "current" if the tests pass.
+# Run testing without writing *.pyc to the __pycache__.
+PYTHONDONTWRITEBYTECODE=1
+(cd "${CASCADE_DEVELOP_DIR}/tests" && python -m pytest -p no:cacheprovider --ihme)
 
 if [ "$?" -eq "0" ]; then
     for softlink in prod dev current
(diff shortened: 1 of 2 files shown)


scripts/automation/install.sh
```diff
@@ -56,7 +56,7 @@ fi
 # Download the Cascade project from github
 
 CASCADE_HOME=/ihme/code/dismod_at
-CASCADE_DEVELOP_DIR=${CASCADE_HOME}/cascade_develop
+CASCADE_DEVELOP_DIR=${CASCADE_HOME}/clone_for_install
```
adolgert: Why change the name?

scripts/automation/install.sh
```diff
@@ -95,8 +95,10 @@ if [ "$?" -ne "0" ]; then
     exit 7
 fi
 
-# Install the virtual environment to be the "current" if the tests pass
-(cd "${CASCADE_DEVELOP_DIR}/tests" && pytest)
+# Install the virtual environment to be the "current" if the tests pass.
+# Run testing without writing *.pyc to the __pycache__.
+PYTHONDONTWRITEBYTECODE=1
+(cd "${CASCADE_DEVELOP_DIR}/tests" && python -m pytest -p no:cacheprovider --ihme)
```
adolgert: If we're going to `chmod -R` it later, why not make cache now?

src/cascade/core/db.py
```diff
@@ -93,6 +93,7 @@ def connection(execution_context=None, database=None):
     connection = ezfuncs.get_connection(database)
     yield connection
     connection.commit()
+    connection.close()
```
adolgert: OMG, how did we miss this? This line was half the point of this context manager.

scripts/automation/install.sh
adolgert: If it doesn't pass unit tests, then this script exits before the chmod, creating a file management problem.

scripts/automation/install.sh
adolgert: Alec is on vacation, so this makes a new directory in order to avoid a directory with bad permissions. Will need to delete the other directory after he returns.

PR #125, opened 2018-12-24
started versioning and made docs about it

adolgert:
New feature that the library now has a version that will appear in the EpiViz-AT log file and in the docs on cascade.readthedocs.io. The version is year.month.day, as agreed with Alec. It requires tagging the library with "git tag -a v18.12.23 -m 'why'". This is explained in the developer documentation.


docs/conf.py
```diff
@@ -60,10 +61,11 @@
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-# The short X.Y version.
-version = "0.1"
 # The full version, including alpha/beta/rc tags.
-release = "0.1"
+release = get_distribution('cascade').version
+# for example take major/minor
```
adolgert: Not sure these comments are necessary. Or should they be more appropriate to this application?

docs/dev/dev.rst
```diff
@@ -9,4 +9,5 @@ Developer Manual
    meid_quota
    testing
    logging
-   dismod_at_shape
\ No newline at end of file
+   dismod_at_shape
```
adolgert: Did you change line 12? Is there extra whitespace?

docs/requirements.txt
```diff
@@ -23,6 +23,7 @@ pytz==2018.5
 PyYAML==3.13
 requests==2.19.1
 scipy==1.1.0
+setuptools_scm
```
adolgert: OMG, we're still using this file.

PR #127, opened 2018-12-26
Feature/locations have algorithms

adolgert:
This makes locations into a full-fledged graph so that we can use graph algorithms on the location graph in order to generate a graph of the work we need to do.

Why? Because I'm working on the drill, and the drill needs to reason about a sequence of locations within the larger location graph. And were I to write my own graph algorithm code, I should be shot.

And the hierarchies library is so very useless. It did nothing for us. I've fallen back to db_queries for its data, so the hierarchies library wasn't even protecting me from SQL queries.

New feature: The location list is now ordered top-to-bottom with a stable sort (lexicographical).
New code feature: The location graph now keeps all the information from the db_queries call as node properties. And, of course, this code composes small, well-tested functions, instead of writing our own graph code.


src/cascade/dismod/serialize.py
```diff
@@ -138,17 +139,21 @@ def make_log_table():
...
-    parent_id = parent.id if parent is not None else np.NaN
-    result = [{"node_name": node.info["location_name_short"], "c_location_id": node.id, "parent": parent_id}]
-    for child in node.level_n_descendants(1):
-        result += rec_build_nodes_table(child, node)
-    return result
+def _parent_of(location_graph, node):
+    try:
+        return next(location_graph.predecessors(node))
+    except StopIteration:
+        return np.nan
```
adolgert: This is used just below. I could make it shorter with a nested comprehension, but this was so much clearer.

src/cascade/dismod/serialize.py
```diff
@@ -138,17 +139,21 @@ def make_log_table():
...
 
 
 def make_node_table(execution_context):
-    locations = get_location_hierarchy_from_gbd(execution_context)
-    table = pd.DataFrame(rec_build_nodes_table(locations.root, None), columns=["node_name", "parent", "c_location_id"])
+    locations = location_hierarchy(execution_context)
+    nodes = list(nx.lexicographical_topological_sort(locations))
+    table = pd.DataFrame(dict(
+        node_name=[locations.nodes[nidx]["location_name"] for nidx in nodes],
```
adolgert: Syntax to get node properties is nice here.

src/cascade/input_data/db/locations.py
```diff
@@ -1,64 +1,115 @@
...
+    """
+    location_df = db_queries.get_location_metadata(
+        location_set_id=35, gbd_round_id=execution_context.parameters.gbd_round_id)
+    G = nx.DiGraph()
+    G.add_nodes_from([(int(row.location_id), row._asdict()) for row in location_df.itertuples()])
+    # GBD encodes the global node as having itself as a parent.
+    G.add_edges_from([(int(row.parent_id), int(row.location_id))
+                      for row in location_df[location_df.location_id != 1].itertuples()])
+    G.graph["root"] = 1  # Global is the root location_id
+    return G
```
adolgert: This function is all of the hierarchies module. The only reason to add both nodes and edges, instead of just edges that imply nodes, is so that we can pick up all of the other columns from the incoming dataframe as node properties.

src/cascade/input_data/db/locations.py
```diff
@@ -1,64 +1,115 @@
...
 
 
 def get_descendents(execution_context, children_only=False, include_parent=False):
```
adolgert: I suspect this function will go away shortly once I have the CascadePlan that tells you what parent and children are. There are only three callers of this function, and they want parent, children, descendants.

PR #128, opened 2018-12-26
Moved script into installed code and made a setup.py entry point for it

adolgert:
Katie asked to work with code programmatically and on the cluster. Given two choices: use our current code or modify the JSON, I recommended modifying the JSON.

This PR documents the epiviz_json.py script and installs it with the code so that it is runnable on the cluster. I augmented documentation (of course). Otherwise no changes of note.


Brianepaulson:
Benefit - so modelers can upload their bundles self-service.

PR #129, opened 2018-12-28
Feature/create ode additional steps

adolgert:
DO NOT MERGE UNTIL NEWER DISMOD-AT THAN 0.7 IS ON THE CLUSTER.
This adds additional ODE integration steps, whether the user specifies them or not.
The modelers and Greg opted for this. It seems safer during early testing to make things a little slower but remove doubt.

It needs dismod that runs. The acceptance test is to run `dmcascade --mvid 266870 --no-upload z.db
Then check that the age steps are in the db and are mentioned in the dismod-at logs.

### DIFF
--- a/src/cascade/executor/epiviz_runner.py
+++ b/src/cascade/executor/epiviz_runner.py
@@ -171,6 +171,12 @@ def add_omega_constraint(model_context, execution_context, sex_id):
 
 
 def add_age_steps(model_context):
+    smallest_step = model_context.parameters.ode_step_size
+    added = compute_age_steps(smallest_step)
+    model_context.parameters.additional_age_steps = added
+
+
+def compute_age_steps(smallest_step):
     """We will add age steps to whatever they have given. The GBD chooses
     age step sizes of 0, 7 days, 28 days, 1 year, 5 years. These look roughly
     like a pattern of multiplying by 4, so 1 week, 4 weeks 16 weeks, 64 weeks.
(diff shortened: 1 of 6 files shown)


src/cascade/executor/epiviz_runner.py
```diff
@@ -169,6 +170,37 @@ def add_omega_constraint(model_context, execution_context, sex_id):
...
 
 
+def add_age_steps(model_context):
+    smallest_step = model_context.parameters.ode_step_size
+    added = compute_age_steps(smallest_step)
+    model_context.parameters.additional_ode_steps = added
+
+
+def compute_age_steps(smallest_step):
+    """We will add age steps to whatever they have given. The GBD chooses
```
alec-deason: "they"? I assume you mean "the modelers"? I think it's better to talk about where the thing we're adding to actually comes from, the settings or the model_context depending on how you want to think about it.

src/cascade/executor/epiviz_runner.py
```diff
@@ -169,6 +170,37 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     model_context.input_data.observations = pd.concat([observations, asdr], ignore_index=True, sort=True)
 
 
+def add_age_steps(model_context):
+    smallest_step = model_context.parameters.ode_step_size
+    added = compute_age_steps(smallest_step)
```
alec-deason: The docstring on "compute_age_steps" implies that these steps will be added to what is specified by the modelers but that's not what's currently happening. What the modeler's specify is ignored and only this value is used. Either I'm confused or there's a bug here.

adolgert: Alec, yes! The current EpiViz-AT form has an increment-decrement button for the ODE step size. It allows you to put in a floating point number, or the little arrows move the value to the next integer. We have a ticket with them asking for this to become an int field. If we use our validation code before it is an int field, then every JSON will fail validation. The modelers specifically asked us to put this set of ODE steps into the code until they could set it themselves. So we are supposed to override whatever we get and ignore the wrong values for now.

alec-deason: Cool

PR #130, opened 2019-01-03
really force files to be world-readable

adolgert:
The log files weren't world-readable. This is a bugfix. For users, this means we were not able to read their error logs, which is a big problem.

The fault is that the code used mkstemp to create a unique file, and mkstemp intentionally sets file permissions to be readable only by the user. I didn't know that behavior. There was no test for readability of the files, and there now is.

One comment: This code avoids using Python's tempfile module by creating its own random string, which is known to be terrible for temporary filenames. This file isn't actually temporary. It's just meant not to crash into other filenames, most of the time. I don't feel so bad about it.

### DIFF
--- a/src/cascade/executor/argument_parser.py
+++ b/src/cascade/executor/argument_parser.py
@@ -7,8 +7,7 @@ from getpass import getuser
 import logging.handlers
 import os
 from pathlib import Path
-from secrets import choice
-import string
+from secrets import token_urlsafe
 import sys
 
 import pkg_resources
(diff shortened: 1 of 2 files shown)


src/cascade/executor/argument_parser.py
```diff
@@ -157,17 +158,18 @@ def _logging_configure_root_log(code_log_dir, level):
...
-        except (OSError, PermissionError) as ose:
-            logging.warning(f"Could not write to file in {user_code_dir} to make a log. {ose}")
+        if not os.access(str(user_code_dir), os.W_OK):
+            logging.warning(f"Could not write to file in {user_code_dir} to make a log")
             return
 
-        os.close(log_file_descriptor)
+        # On collision, this will fail to write a code log.
+        # Not using tempfile because it insists on secure mode flags.
+        uniquish = ''.join(choice(string.ascii_letters) for _ in range(8))
```
alec-deason: Why not the token_urlsafe function? That should also make filename safe sequences and is simpler

PR #131, opened 2019-01-04
Constrain column order when freezing data and fix sex_id/x_sex confusion in country covariates

alec-deason:
This fixes two unrelated bugs.

One is that the country covariate setup code was assuming that bundle data used the x_sex convention to represent sex on data rather than GBD's sex_id convention.

The other is that the code that freezes bundle, csmr and asdr data was not careful enough with column order which made it possible for data to end up in the wrong column in the database and thus disaster. Also the code which calculates standard error from upper and lower bounds was not checking that the bounds make sense, but does now.

### DIFF
--- a/src/cascade/input_data/configuration/construct_country.py
+++ b/src/cascade/input_data/configuration/construct_country.py
@@ -427,6 +427,8 @@ def get_covariate_data_by_sex(covariates):
         covariates_both["avg_time"] = covariates_both["avg_time_x"]
         covariates_by_sex[BOTH] = covariates_both
     else:
-        raise ValueError(f"Unexpected values for sex_id in covariates data.  Expected 3 or (1,2), found {sex_values}")
+        raise ValueError(
+            f"Unexpected values for sex_id in covariates data.  Expected (3,4) or (1,2), found {sex_values}"
+        )
 
     return covariates_by_sex
--- a/src/cascade/stats/estimation.py
+++ b/src/cascade/stats/estimation.py
@@ -20,8 +20,9 @@ def meas_bounds_to_stdev(df):
     Replace any zero values with :math:`10^{-9}`.
     """
     MATHLOG.debug("Assigning standard error from measured upper and lower.")
-    if np.any(df.meas_lower > df.meas_value) or np.any(df.meas_value > df.meas_upper):
-        raise ValueError(f"Found data where meas_lower > meas_value or meas_upper < meas_value")
+    bad_rows = np.sum((df.meas_lower > df.meas_value) | (df.meas_value > df.meas_upper))
+    if bad_rows > 0:
+        raise ValueError(f"Found data where meas_lower > meas_value or meas_upper < meas_value on {bad_rows} rows")
     df["standard_error"] = (df.meas_upper - df.meas_lower) / (2 * 1.96)
     df["standard_error"] = df.standard_error.replace({0: 1e-9})
     df = df.rename(columns={"meas_value": "mean"})
--- a/tests/stats/test_estimation.py
+++ b/tests/stats/test_estimation.py
@@ -1,0 +1,12 @@
+import pytest
+
+import pandas as pd
+
+from cascade.stats.estimation import meas_bounds_to_stdev
+
+
+def test_meas_bounds_to_stdev__bad_bounds():
+    df = pd.DataFrame({"meas_lower": [0, 0, 1, 1], "meas_value": [0, -0.01, 1, 1], "meas_upper": [-10, 0, 1, 10]})
+
+    with pytest.raises(ValueError):
+        meas_bounds_to_stdev(df)


src/cascade/input_data/configuration/construct_country.py
```diff
@@ -170,7 +171,7 @@ def covariate_to_measurements_nearest_favoring_same_year(measurements, sex, cova
             zip(
                 covariates[["age_lower", "age_upper"]].mean(axis=1) / 240,
                 covariates[["time_lower", "time_upper"]].mean(axis=1),
-                covariates["x_sex"],
+                covariates["sex_id"],
```
adolgert: This seems unlikely to work. Is this function even used? The problem is that this KDTree thing relies on having data points assigned to -0.5, 0, 0.5. If they are assigned to 1, 2, 3, they are distributed spatially in a different way. Really changes what this does.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -202,14 +202,12 @@ def convert_gbd_ids_to_dismod_values(with_ids_df, age_groups_df):
         pd.DataFrame: New pd.DataFrame with four added columns and in the same
             order as the input dataset.
     """
-    sex_df = pd.DataFrame(dict(x_sex=[-0.5, 0, 0.5], sex_id=[2, 3, 1]))
```
adolgert: Aw, I was trying to see how many places we could put this information. Figure we were up to 4 or 5.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -311,13 +309,12 @@ def get_measurement_data_by_sex(measurements):
...
     Returns:
-        dict: possible sex keys (-0.5, 0, 0.5) and measurement data as values
+        dict: possible sex keys (1, 2, 3, 4) and measurement data as values
     """
     measurements_by_sex = {}
 
-    for sex in (FEMALE, MALE, BOTH):
-
-        measurements_sex = measurements[measurements["x_sex"] == sex]
+    for sex in (FEMALE, MALE, BOTH, UNDEFINED):
```
adolgert: This is effectively a new feature.

src/cascade/input_data/configuration/construct_country.py
```diff
@@ -430,6 +427,6 @@ def get_covariate_data_by_sex(covariates):
         covariates_both["avg_time"] = covariates_both["avg_time_x"]
         covariates_by_sex[BOTH] = covariates_both
     else:
-        raise ValueError(f"Unexpected values for x_sex in covariates data.  Expected 3 or (1,2), found {sex_values}")
+        raise ValueError(f"Unexpected values for sex_id in covariates data.  Expected 3 or (1,2), found {sex_values}")
```
adolgert: "Expected (3,4) or (1,2)."

src/cascade/input_data/db/asdr.py
```diff
@@ -95,9 +83,23 @@ def _upload_asdr_data_to_tier_3(execution_context, cursor, model_version_id, asd
         "age_group_years_start": "age_lower",
         "age_group_years_end": "age_upper"
     })
-
     asdr_data = asdr_data.merge(age_group_data, how="left", on="age_group_id")
     asdr_data = asdr_data.where(pd.notnull(asdr_data), None)
```
adolgert: This is an interesting line. It implies some of the asdr data could be null? I should hope not. Not for this PR, I guess, but I wish I knew why this was here. We could warn when this happens.

src/cascade/input_data/db/asdr.py
```diff
@@ -95,9 +83,23 @@ def _upload_asdr_data_to_tier_3(execution_context, cursor, model_version_id, asd
...
+        "age_group_id",
+        "mean",
+        "upper",
+        "lower",
+        "age_upper",
+        "age_lower",
+    ]
+
+    asdr_data = asdr_data[ordered_cols]
+
```
adolgert: What if we used Pandas to_sql? Would that sort out the columns automatically for the upload? Or we could upload using the technique where you use todict() instead of tolist(). That way each row has keys and gets aligned correctly. Or make sqlalchemy metadata.

Walking back, would Pandas nicely order the upload for us?

src/cascade/input_data/db/bundle.py
```diff
@@ -13,6 +13,41 @@
 
 CODELOG, MATHLOG = getLoggers(__name__)
 
+BUNDLE_COLUMNS = [
```
adolgert: This method ensures the two are consistent. Again, would Pandas do this?

src/cascade/input_data/db/csmr.py
```diff
@@ -83,7 +83,7 @@ def get_csmr_data(execution_context):
 
     csmr = csmr[csmr["val"].notnull()]
 
-    csmr.rename(columns={"val": "mean"})
+    csmr = csmr.rename(columns={"val": "mean"})
```
adolgert: How did this get by us? That was bad. And does fixing this mean we expect "val" later? What's downstream?

src/cascade/stats/estimation.py
```diff
@@ -18,6 +20,8 @@ def meas_bounds_to_stdev(df):
     Replace any zero values with :math:`10^{-9}`.
     """
     MATHLOG.debug("Assigning standard error from measured upper and lower.")
+    if np.any(df.meas_lower > df.meas_value) or np.any(df.meas_value > df.meas_upper):
+        raise ValueError(f"Found data where meas_lower > meas_value or meas_upper < meas_value")
```
adolgert: This would have helped earlier. If you have time, and sanity enough, it would be cool to report how many lines are out of whack. Not necessary.

src/cascade/input_data/configuration/construct_country.py
adolgert: I'm walking this back. It might still work. The idea was that a data point at x_sex=0.0 would be close to 0.5 and -0.5, if they were missing. But a point at 3 is still closest to missing values at 1 and 2 if they are gone.

src/cascade/input_data/configuration/construct_country.py
alec-deason: Yeah, I wasn't sure what to do. So I tried not to make a decision.

src/cascade/input_data/configuration/construct_country.py
alec-deason: Yes

src/cascade/input_data/db/asdr.py
alec-deason: Either Simon or Natalie had a reason, but I can't remember what it was.

src/cascade/input_data/configuration/construct_country.py
adolgert: It's a good choice. The modelers are clear that internally we should treat undefined as both.

src/cascade/input_data/db/csmr.py
alec-deason: Yeah, that's pretty scary. It was getting implicitly renamed int the freeze-unfreeze operation because :bomb:

src/cascade/stats/estimation.py
alec-deason: Yeah, I agree. I was trying to do the quick thing but I guess there isn't really a rush. I'll improve the message.

src/cascade/input_data/db/bundle.py
alec-deason: Probably. We ended up here because I was just porting what the old importer does. I'm going to say we don't pursue that in this PR and do it in a more complete cleanup later. Mostly because I don't know panda's SQL layer well and I'd want to understand it better.

src/cascade/input_data/configuration/construct_country.py
alec-deason: Sorry to spoil your fun

src/cascade/input_data/db/bundle.py
adolgert: Got it. OK. Visual inspection on those lists was good for me.

src/cascade/input_data/configuration/construct_country.py
alec-deason: It is not used. Should I just dump it?

PR #134, opened 2019-01-07
Feature/warnings are errors during unit testing

adolgert:
1. For unit tests, this turns on a flag to make all warnings errors. Warnings are a particular class of messages in Python, coming from the warnings module, that are used by many libraries to indicate future incompatibility or current suspicious behavior. This also suppresses certain known warnings. The result is that unit tests now pass "full green," instead of yellow, as before.
2. Unit testing on the Mac failed because some of the tests intentionally kill Dismod in such a way that the operating system offers to tell Apple that the application crashed (with a SIGINT). This adds a fixture that can be used to turn _off_ those tests, called `--signals`.
3. Our wrapper script for Dismod-AT let's you call Dismod-AT from the command line both on the cluster (where it is installed under Singularity) or on your laptop (using Docker) with the same command. The wrapper was failing on a laptop with a different version of Bash (my Mac), so there is a one-line fix for that.

### DIFF
--- a/scripts/dmdismod
+++ b/scripts/dmdismod
@@ -4,7 +4,7 @@ if [ -e "${SINGULARITY}" ]
 then
     exec singularity exec "${SINGULARITY}" /home/root/prefix/dismod_at/bin/dismod_at $*
 else
-    TARGET_DIR=$(readlink -e $(dirname $1))
+    TARGET_DIR="$( cd "$(dirname $1)" ; pwd -P )"
     TARGET_DB=$(basename $1)
     shift
     set -x
--- a/src/cascade/input_data/emr.py
+++ b/src/cascade/input_data/emr.py
@@ -1,4 +1,5 @@
 from textwrap import dedent
+import warnings
 
 from scipy.interpolate import InterpolatedUnivariateSpline, LSQBivariateSpline
 
@@ -83,12 +84,14 @@ def _make_interpolators(csmr):
     """
     mean = {}
     stderr = {}
-    mean["both"] = LSQBivariateSpline(
-        csmr.age, csmr.time, csmr["mean"], sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
-    )
-    stderr["both"] = LSQBivariateSpline(
-        csmr.age, csmr.time, csmr.standard_error, sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
-    )
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", module="scipy.interpolate")
+        mean["both"] = LSQBivariateSpline(
+            csmr.age, csmr.time, csmr["mean"], sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
+        )
+        stderr["both"] = LSQBivariateSpline(
+            csmr.age, csmr.time, csmr.standard_error, sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
+        )
 
     csmr_by_age = csmr.sort_values("age").groupby("age").mean().reset_index()
     mean["age"] = InterpolatedUnivariateSpline(csmr_by_age.age, csmr_by_age["mean"], k=1)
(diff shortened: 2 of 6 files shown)


scripts/dmdismod
```diff
@@ -4,7 +4,7 @@ if [ -e "${SINGULARITY}" ]
 then
     exec singularity exec "${SINGULARITY}" /home/root/prefix/dismod_at/bin/dismod_at $*
 else
-    TARGET_DIR=$(readlink -e $(dirname $1))
+    TARGET_DIR="$( cd "$(dirname $1)" ; pwd -P )"
```
alec-deason: Greg will be happy

src/cascade/input_data/emr.py
```diff
@@ -83,12 +84,14 @@ def _make_interpolators(csmr):
...
     mean = {}
     stderr = {}
-    mean["both"] = LSQBivariateSpline(
-        csmr.age, csmr.time, csmr["mean"], sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
-    )
-    stderr["both"] = LSQBivariateSpline(
-        csmr.age, csmr.time, csmr.standard_error, sorted(csmr.age.unique()), sorted(csmr.time.unique()), kx=1, ky=1
-    )
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", module="scipy.interpolate")
```
alec-deason: Is this always safe? We don't need to filter on some more granular value?

src/cascade/input_data/emr.py
```diff
@@ -105,7 +108,7 @@ def _emr_from_sex_and_node_specific_csmr_and_prevalence(csmr, prevalence):
     mean_interp, stderr_interp = _make_interpolators(csmr)
 
     emr = prevalence[["age_lower", "age_upper", "time_lower", "time_upper", "sex_id", "node_id", "density", "weight"]]
-    emr["measure"] = "mtexcess"
+    emr = emr.assign(measure="mtexcess")
```
alec-deason: Is this cleaner than using `.copy()` and keeping the slice assignment syntax? Maybe it is...

tests/executor/test_dismod_runner.py
```diff
@@ -45,7 +45,7 @@ def test_dmchat_low_priority(dmchat, caplog):
     assert err == "".join(["err" + os.linesep] * 2)
 
 
-def test_dmchat_nonzero(dmchat):
+def test_dmchat_nonzero(dmchat, signals):
```
alec-deason: I like this pattern.

tests/model/test_operations.py
```diff
@@ -54,8 +54,8 @@ def make_grid_priors(ages, times, grid_priors, hyper_priors, smooth_id=7):
                     hyper_priors["value"], hyper_priors["dage"], hyper_priors["dtime"], 0],
     ))
     # Make the last one const. If tests want to make a point const, use prior_id=6.
-    prior.loc[6]["lower"] = 0.5
-    prior.loc[6]["upper"] = 0.5
+    prior.loc[6, "lower"] = 0.5
```
alec-deason: Oh, gross. That shouldn't have happened.

tox.ini
```diff
@@ -3,6 +3,11 @@ envlist = py36,docs
 [pytest]
 xfail_strict = true
 testpaths = tests
+filterwarnings =
+    error::Warning
+    ignore:invalid escape sequence:DeprecationWarning
```
alec-deason: What throws this one?

src/cascade/input_data/emr.py
adolgert: Changed to be very specific.

src/cascade/input_data/emr.py
adolgert: Erm, this was one of the ones that was "assignment to a slice." I reread the Pandas docs a month or two ago, and they said "assign" was the new thing to use. I'm trying to be obedient to norms (where I can).

tests/executor/test_dismod_runner.py
adolgert: It's one of a few possible patterns. It's working for me so far. I think there may be a third flag, for "dismod is present on this machine".

tests/model/test_operations.py
adolgert: In test code, that I wrote. Yup. No big deal.

tox.ini
adolgert: I should comment those. Can I comment each line in this file? The invalid escape sequence is for using Latex inside a docstring that isn't a "raw" docstring, with the r in front. Unfortunately, networkx does this, but I think some others do too. It's still a good library. I swear. Used it for years.

alec-deason: I'm pro-networkx. Don't need to convince me.

src/cascade/input_data/emr.py
alec-deason: Cool

PR #137, opened 2019-01-14
Feature/load to tier three as needed

adolgert:
1. Renamed descendent to descendant. This was the most important contribution of this PR.
2. ASDR and CSMR now check, not that any data is in tier 3, but that any data from the locations in this location and its descendants, are in tier 3. Then it conditionally loads data from locations which are missing data.
3. I also deleted blank lines, because I'm a professional, and professionals hate whitespace.

### DIFF
--- a/src/cascade/input_data/db/asdr.py
+++ b/src/cascade/input_data/db/asdr.py
@@ -4,7 +4,7 @@ import pandas as pd
 
 from cascade.core.db import cursor, db_queries
 from cascade.input_data.db import AGE_GROUP_SET_ID
-from cascade.input_data.db.locations import get_descendents
+from cascade.input_data.db.locations import get_descendants
 
 
 from cascade.core.log import getLoggers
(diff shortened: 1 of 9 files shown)


src/cascade/input_data/db/asdr.py
```diff
@@ -111,24 +105,24 @@ def load_asdr_to_t3(execution_context) -> bool:
...
+    locations_with_asdr_in_t3 = _asdr_in_t3(execution_context)
+    missing_from_t3 = set(location_and_children) - set(locations_with_asdr_in_t3)
+    if missing_from_t3:
         CODELOG.info(
             f"""Uploading asdr data for model_version_id
             {model_version_id} on '{database}'"""
         )
-
-        asdr_data = get_asdr_data(execution_context)
+        asdr_data = get_asdr_data(execution_context, list(missing_from_t3))
```
alec-deason: Just because the underlying API barfs on sets, our API should accept any iterable

PR #139, opened 2019-01-17
Feature/interactive classes

adolgert:
It's time to take a look at this branch with core classes for Model, Session, DismodGroups, Var, and SmoothGrid. Those are the central pieces.

New features!

1. We can check and set scale vars, which determine the relative importance of different terms in the problem and should help explain some modeler questions.
2. We can check and set the start_var, which is the initial place the optimizer tries to find a solution.
3. The weigt grids are defined for the first time. We haven't set them before, and this introduces a method for that.
4. All options are set to explicit defaults.
5. You can predict without doing a model fit (at least without doing it in a way you'd notice).

None of these features are available to users yet. This PR creates the core functions and opens them up to review. Later PRs will

1. Document an API with examples
2. Explore and test more of the API: use of random effects, use of covariates.
3. Create a better interface to the prior setting and getting.
4. Change the IHME db retrieval to retrieve in a shape in order to set it for these models.


src/cascade/dismod/db/wrapper.py
```diff
@@ -17,14 +17,15 @@
...
+from cascade.dismod.constants import DensityEnum
 from cascade.dismod.db import DismodFileError
 
-from cascade.core.log import getLoggers
+from cascade.core import getLoggers
 CODELOG, MATHLOG = getLoggers(__name__)
 
 
-def _get_engine(file_path):
+def get_engine(file_path):
```
alec-deason: Yes, this is as it should be. Probably makes sense to have a factory for the common `DismodFile(get_engine(Path("...")))` case too.

src/cascade/model/priors.py
```diff
@@ -213,3 +213,42 @@ def _parameters(self):
 ZERO = Uniform(0, 0, 0, name="constrain_to_zero")
 ZERO_TO_ONE = Uniform(0, 1, 0.1, name="uniform_zero_to_one")
 MINUS_ONE_TO_ONE = Uniform(-1, 1, 0, name="uniform_negative_one_to_one")
+
+
+DENSITY_ID_TO_PRIOR = {
```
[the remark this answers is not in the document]
adolgert: Yeah, I was wondering about that. I leaned towards a mapping because it was a relationship among existing items, not the definition of a _class_ of items. I'll stare at this again.

src/cascade/model/dismod_groups.py
```diff
@@ -0,0 +1,60 @@
...
+    """Dismod-AT documentation discusses five kinds of groups of
+    model variables (https://bradbell.github.io/dismod_at/doc/model_variables.htm).
+    This class represents that grouping as a set of dictionaries, where the
+    values can either be SmoothGrids or RandomFields or whatever is classified
+    according to groups of model variables.
+
+     * Rate key is rate as a string (iota, rho, chi, omega, pini)
+     * Random effect key is tuple (rate, location_id), where None is all.
+     * Alpha key is (covariate, rate), both as strings.
+     * Beta key is (covariate, integrand), both as strings.
```
alec-deason: This makes me a little nervous. Especially since the Beta key and the Gamma key are the same. But I see that you're storing the groups separately...

My impulse is to ask for something like:
```
my_dismod_groups[Gamma("covariate", "integrand")] = whatever
```
But that adds complexity over what you have and I don't know it really gets us anything. Maybe this:

```
my_dismod_groups.gamma[("covariate", "integrand")] = somethig
```

alec-deason: You could still get a clean iter over the different groups if that's why you're choosing the nested dict like structure in the first place. Would be a short implementation.

src/cascade/model/model.py
```diff
@@ -0,0 +1,115 @@
...
+    and ``time_id`` for Dismod-AT.
+    """
+    def __init__(self, nonzero_rates, parent_location, child_location, covariates=None, weights=None):
+        """
+        >>> locations = location_hierarchy(execution_context)
+        >>> m = Model(["chi", "omega", "iota"], 6, locations)
+        """
+        super().__init__()
+        self.nonzero_rates = nonzero_rates
+        self.location_id = parent_location
```
alec-deason: `parent_location`?

src/cascade/model/model.py
```diff
@@ -0,0 +1,115 @@
...
+        # Covariates are here because their reference values are part of
+        # the model. Even though avgint and data use them, a model is always
+        # written before the avgint and data are written.
+        self.covariates = covariates if covariates else list()
+        # There are always four weights, constant, susceptible,
+        # with_condition, and total.
+        if weights:
+            self.weights = weights
+        else:
+            self.weights = dict()
```
alec-deason: If there are always four, should this be constrained to a type that knows that:
```
class Weights:
    __slots__ = ["constant", "susceptible", "with_condition"]
...
```
Then you could push some of the checks in `_ensure_weights` down into that

src/cascade/model/dismod_groups.py
adolgert: This class is pretty obscure to stare at. It ends up giving `my_dismod_groups.gamma[("covariate", "integrand")] = something`. I'd like to add checking for the keys, so that each of rate, re, alpha, beta, gamma is a dictionary that checks its keys. The pattern in model_reader ends up OK as a nested for loop. I'm not complaining about that. That's internal code, written only a few times, and can be harder to write without penalty. I want this good for the user.

I was trying to leave open making this an object with members, so groups.rate.iota and groups.alpha.traffic.iota. There's a little challenge in using numbers for locations, but we could prefix an underscore.

src/cascade/model/model.py
adolgert: Not sure what's confusing. There's always a parent location, right?

src/cascade/model/model.py
adolgert: Lots of checks are missing in this PR. I was very intentionally light on checks at the API interface until we agree on the interface. And I'm the one making up this rule about the four weights. The modelers should sign off on it.

tests/dismod/db/test_wrapper.py
```diff
@@ -284,10 +284,34 @@ def test_write_covariate_column__bad_name(base_file, dummy_data_row):
...
     dm_file.data = dummy_data_row
     dm_file.flush()
 
-    dm_file2 = DismodFile(_get_engine(None))
+    dm_file2 = DismodFile(get_engine(None))
     table = dm_file2.data
     assert "x_sex" not in table
+
+
+def test_write_empty_table(tmp_path):
```
adolgert: I put in this unit test to find the bug in serialization that was fixed above.

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+    >>> assert isinstance(atg[0, 1990], pd.DataFrame)
+
+    If the column has the same name as a function (mean), then access it
+    with getitem,
+
+    >>> atg[:, :]["mean"] = [5.9]
+
+    """
+    def __init__(self, age_time_grid, columns, count=1):
+        assert isinstance(columns[0], str)
```
alec-deason: If you're going to do it, do all the columns not just one

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+    >>> assert (atg[20, 2000].weight == 195).all()
+    >>> assert isinstance(atg[0, 1990], pd.DataFrame)
+
+    If the column has the same name as a function (mean), then access it
+    with getitem,
+
+    >>> atg[:, :]["mean"] = [5.9]
+
+    """
+    def __init__(self, age_time_grid, columns, count=1):
```
alec-deason: Is there a reason for age_time_grid to be a tuple of lists rather than two arguments? It feels like that reduces clarity and makes a whole class of bugs possible that wouldn't be otherwise.

src/cascade/model/age_time_grid.py
alec-deason: Oh, I see, you're also allowing this to be a single string... I think I would group this in with the stuff on line 46 for clarity. I know that moves the precondition check away from the top of the function but I think that's ok...

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+            self.ages = np.array(age_time_grid[0], dtype=np.float)
+            self.times = np.array(age_time_grid[1], dtype=np.float)
+        except TypeError:
+            raise TypeError(f"Ages and times should be arrays of floats {age_time_grid}.")
+        if isinstance(columns, str):
+            columns = [columns]
+        try:
+            self.columns = list(columns)
+        except TypeError:
+            raise TypeError(f"Columns should be an iterable of strings. {columns}")
```
alec-deason: "Columns should be either a string or an iterable of strings"

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+        for col_is_str in self.columns:
+            if not isinstance(col_is_str, str):
+                raise TypeError(f"Columns should be iterable of strings. {col_is_str}")
```
alec-deason: Between this and the check on 46 you've redone the check on 37

alec-deason: Also the error message should describe the actual type constraint, not the particular part of it which was violated in this check, I think. Same as my comment above.

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+                age=[nan],
+                time=[nan],
+                idx=list(range(count)),
+            ))
+            mulstd_df = mulstd_df.assign(**{new_col: nan for new_col in columns})
+            self._mulstd[kind.name] = mulstd_df
+
+    @property
+    def mulstd(self):
+        return self._mulstd
```
alec-deason: Is the intention that this be imutable? Might want to return a copy.

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+
+    >>> atg = AgeTimeGrid(([0, 10, 20], [1990, 2000, 2010]), ["height", "weight"])
+    >>> atg[:, :] = [6.1, 195]
+    >>> atg[:, :].height = [5.9]
+    >>> atg[10, 2000] = [5.7, 180]
+    >>> atg[5:17, 1980:1990].weight = 125
+    >>> assert (atg[20, 2000].weight == 195).all()
+    >>> assert isinstance(atg[0, 1990], pd.DataFrame)
+
+    If the column has the same name as a function (mean), then access it
```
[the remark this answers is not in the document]
alec-deason: Also, is this even true? `atg[:, :]` will always return a DataFrame (I think) and indexing a method name out of a dataframe doesn't work this way.

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+            at_range.append([start - GRID_SNAP_DISTANCE, stop + GRID_SNAP_DISTANCE])
+        ages = self.ages[(at_range[0][0] <= self.ages) & (self.ages <= at_range[0][1])]
+        times = self.times[(at_range[1][0] <= self.times) & (self.times <= at_range[1][1])]
+        if len(ages) == 0:
+            raise ValueError(f"No ages within range {at_range[0]}")
+        if len(times) == 0:
+            raise ValueError(f"No times within range {at_range[1]}")
+        self.grid.loc[np.in1d(self.grid.age, ages) & np.in1d(self.grid.time, times), self.columns] = value
+
+    def __len__(self):
```
alec-deason: Is this really a meaningful length for this object? My intuition that "length" on multi-dimensional structures is the length of the first axis but that's not what you calculate here. Maybe this structure just shouldn't have len?

src/cascade/model/age_time_grid.py
```diff
@@ -0,0 +1,153 @@
...
+        mulstd_cnt = sum(not df[self.columns].dropna(how="all").empty for df in self._mulstd.values())
+        return self.ages.shape[0] * self.times.shape[0] + mulstd_cnt
+
+    def __str__(self):
```
alec-deason: You should, I think, either do __str__ and __repr__ or just __repr__ but not just __str__. __str__ falls back to the class's __repr__ but __repr__ falls back to Object's repr which isn't useful.

```
>>> class Test:
...   def __str__(self):
...     return "test"
...
>>> t = Test()
>>> str(t)
'test'
>>> repr(t)
'<__main__.Test object at 0x7ff157a75828>'
>>> class Test:
...   def __repr__(self):
...     return "test"
...
>>> t = Test()
>>> str(t)
'test'
>>> repr(t)
'test'
```

src/cascade/model/dismod_groups.py
alec-deason: Oh, it does have the getattr implementation. That's fine then. Sorry I overlooked that.

src/cascade/model/model.py
alec-deason: Sorry, what I meant was: Should it be `self.parent_location` rather than `self.location_id` I think that would be more clear.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+
+    Returns:
+        DismodGroups: With means for everything.
+    """
+    var_name = f"{which}_var"
+    table = getattr(dismod_file, var_name)
+    return _assign_from_var_ids(table, var_ids, partial(_read_vars_one_field, name=var_name))
+
+
+def _assign_from_var_ids(table, var_ids, assignment):
```
alec-deason: I find the name `assignment` to be very confusing. I actually spent quite a while looking for where it was defined when I saw it down below. Then I realized it was an argument to the function. I think I would have been less confused with the generic but conventional `func`. Also is there a reason not to call this `_map_over_var_ids` or something? It actually doesn't do any assignment, it returns a new value.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+        raise AttributeError(f"Dismod file has no data in table {table.columns}.")
+    var_groups = DismodGroups()
+    for group_name, group in var_ids.items():
+        for key, value in group.items():
+            var_groups[group_name][key] = assignment(table, value)
+    return var_groups
+
+
+def _read_vars_one_field(table, id_draw, name):
+    """Read one rate or randome effect's values.
```
alec-deason: randome -> random

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+    Returns:
+        A new Var containing the values at these ages and times.
+    """
+    id_column = f"{name}_id"
+    var_column = f"{name}_value"
+    table = table.reset_index(drop=True)
+    with_var = id_draw.grid.merge(table, left_on="var_id", right_on=id_column, how="left")
+
+    vals = Var((id_draw.ages, id_draw.times))
+    vals.grid = vals.grid.drop(columns=["mean"]) \
```
alec-deason: I think this is a fine way to do this but I believe it would be clearer to me as index alignment followed by assignment rather than a merge surrounded by drop and rename

src/cascade/model/age_time_grid.py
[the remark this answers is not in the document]
adolgert:
```python
>>> df = pd.DataFrame(dict(a=[1,2,3], mean=[4,5,6]))
>>> df.a
0    1
1    2
2    3
Name: a, dtype: int64
>>> df.mean
<bound method DataFrame.mean of    a  mean
0  1     4
1  2     5
2  3     6>
>>> df["mean"]
0    4
1    5
2    6
Name: mean, dtype: int64
```
I don't like this. I'm torn between using "mean" because it's everywhere in Dismod's documentation and because, for Priors, it's actually the correct term, and using another name. Next guess is "value", but that's like the "values" property.

src/cascade/model/age_time_grid.py
adolgert: The reason... I was thinking they are always together, but they truly aren't, and it's starting to feel stupid. Let's change it to ages and times as separate arguments.

src/cascade/model/age_time_grid.py
adolgert: Maybe this shouldn't be len. There's one really important length here, the number of model variables. I put a `count()` method on the `DismodGroups` which then adds up the lengths of all contained items. I could make it `variable_count()`?

src/cascade/model/age_time_grid.py
adolgert: OK. How embarrassing.

src/cascade/model/model_reader.py
adolgert: Yup. Yup. Yup. Highly misleading.

src/cascade/model/age_time_grid.py
adolgert: What's in the dict has to be mutable. I just didn't want people assigning to the mulstd property... too much?

src/cascade/model/model_reader.py
adolgert: Ahah! I can do this with a straight-up assign statement!

src/cascade/model/age_time_grid.py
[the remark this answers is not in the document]
alec-deason: Oh... When you say "has the same name as a function" you don't mean that it will execute that function but that you will populate the data with the results of that function under the expected name? That's less weird than what I thought. But is that happening somewhere? I don't see it. Or am I just confused?

src/cascade/model/age_time_grid.py
alec-deason: No, that's fine. It's just that I usually use this pattern to produce immutability, both superficial and deep, but if that's not what you want then this is fine.

src/cascade/model/age_time_grid.py
alec-deason: I think I like that better. `len` has ambiguous semantics for complex structures. And I think it's only worth dealing with that if you have to.

src/cascade/model/age_time_grid.py
alec-deason: Meh. I have to look this up every time I do it.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+                group_name, key = inverted_smooth[(smooth_id, node_id)]
+                var_ids[group_name][key] = _add_one_field_to_vars(re_grid_df, age, time)
+
+    if var_ids.count() != len(dismod_file.var):
+        MATHLOG.error(f"Found {var_ids.count()} of {len(dismod_file.var)} vars in db file.")
+
+    return rename_node_to_location(dismod_file.node, var_ids)
+
+
+def _add_one_field_to_vars(sub_grid_df, age, time):
```
alec-deason: Similar to a comment above. This extracts data from a data frame, it doesn't add anything so the name is confusing.

src/cascade/model/age_time_grid.py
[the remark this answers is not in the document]
adolgert: It's not going wrong anywhere. It's a very common pitfall in Pandas, so I mentioned it in the user docs. I don't know whether it's appropriate to add that as an aside, but it's a guess.

src/cascade/model/age_time_grid.py
adolgert: I moved these around to check things once. Thanks for the feedback.

src/cascade/model/age_time_grid.py
adolgert: Let's deal with it now.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+            if group_name == "random_effect":
+                child_node = key[1]
+                inverted_smooth[(smooth_value, child_node)] = (group_name, key)
+            else:
+                inverted_smooth[(smooth_value, parent_node)] = (group_name, key)
+
+    return inverted_smooth
+
+
+def _read_rate_smooths(child_node, nslist_pair_table, rate_table, smooths):
```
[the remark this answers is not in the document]
adolgert: I like this observation! Testing is easier when it returns something. Some of them modify several tables. I'll go through the lot and see if I can clarify...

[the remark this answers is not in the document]
adolgert: And, btw, I have tests to add for all of the functions in this file that read from the dismod_file. That seems to be where the risk is.

src/cascade/model/age_time_grid.py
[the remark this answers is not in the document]
alec-deason: Hmm. Actually rereading what you wrote, you clearly didn't mean what I thought you meant. You are describing what to do not describing a behavior that the system will have... I'm not sure. It wasn't clear to me but that is probably my fault.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+    for rate_row in rate_table.itertuples():
+        if rate_row.parent_smooth_id is not None and not isnan(rate_row.parent_smooth_id):
+            smooths.rate[rate_row.rate_name] = int(rate_row.parent_smooth_id)
+        if rate_row.child_smooth_id is not None and not isnan(rate_row.child_smooth_id):
+            # Random effects can have children with different fields but same smoothing.
+            for child in child_node:
+                smooths.random_effect[(rate_row.rate_name, child)] = int(rate_row.child_smooth_id)
```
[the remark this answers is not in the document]
adolgert: Damned Pandas.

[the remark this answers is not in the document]
alec-deason: Arguably numpy's fault for not having a nullable integer type

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+    """Get ``node_id`` for parent location."""
+    return int(dismod_file.option.query("option_name == 'parent_node_id'").option_value)
+
+
+def read_child_nodes(dismod_file, parent_node):
+    """Get node ids for child locations."""
+    return dismod_file.node[dismod_file.node.parent == parent_node].node_id.values
+
+
+def rename_node_to_location(node_table, var_ids):
```
alec-deason: I assume the inverse of this lives in the writer? That makes sense though it's a little odd to have them seperated.

src/cascade/model/model_reader.py
```diff
@@ -0,0 +1,313 @@
...
+    """Given a DismodGroups based on node_id, change keys
+    to be based on location_id"""
+    node_to_location = DismodGroups()
+    for group_name, group in var_ids.items():
+        for key, smooth_value in group.items():
+            if group_name == "random_effect":
```
alec-deason: Hmm. In the old model this needed to happen in a lot of other places. What about the data? That also has node/locations in it.

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+       to ``location_id``.
+     * Rates and integrands are always in the same order.
+    """
+
+    def __init__(self, session, extremal_age_time):
+        """
+
+        Args:
+            session (Session): The Dismod-AT Session into which this writes.
+            extremal_age_time (Tuple(Set(int),Set(int))): a set of ages and
```
alec-deason: Can we really not defer this so the user doesn't have to think about it? This must be available when write_ages_and_times is called, right? It could be collected at that point?

alec-deason: Or the intention here is that there will be the need to add int extremes beyond what's in the model and data?

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+                passes in the extremal values.
+        """
+        self._session = session
+        self._dismod_file = session.dismod_file
+        if extremal_age_time:
+            self._ages = np.array(list(extremal_age_time[0]), dtype=np.float)
+            self._times = np.array(list(extremal_age_time[1]), dtype=np.float)
+        else:
+            self._ages = np.zeros((0,), dtype=np.float)
+            self._times = np.zeros((0,), dtype=np.float)
```
alec-deason: Do you meant `np.empty`? I'm not sure which is weirder, an array of length zero initialized with zeros (even though that means nothing since there are no elements) or an array of length zero uninitialized (even though that means nothing since there are no elements). I think I like `np.empty` better?

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+        self._rate_rows = list()  # List of dictionaries for rates.
```
alec-deason: If you're going to do it, do it. Type hinting for instance variables: https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#classes

src/cascade/model/model_reader.py
adolgert: It lives in the session. The model knows about children but not all the locations. But, yeah, separate... These readers/writers don't quite have the right form if they are done such different ways. I think it's a refactor, but I don't know yet to what.

src/cascade/model/model_reader.py
adolgert: There's going to be a little appy() in the data that switches location to node_id. Yup. One line.

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+        the tables by telling me the rate count and child random effect count."""
+        self._children = children
+        iota_case = "pos" if "iota" in nonzero_rates else "zero"
+        rho_case = "pos" if "rho" in nonzero_rates else "zero"
+        self._session.set_option(rate_case=f"iota_{iota_case}_rho_{rho_case}")
+
+    def write_ages_and_times(self, ages, times):
+        """
+        Collects all of the ages and times because they are ordered
+        and indexed by Dismod's file format. Can be called multiple times.
```
alec-deason: Hmm, I see. This doesn't actually know the extremes because it may be called more than once. So you must either calculate the extremes outside of this or defer writing of all values that depend on age_id and time_id until you know this has been called as many times as it's going to be. Which means this whole system would need to be written as setup->accumulate->flush and that's a bit ugly. Hm...

src/cascade/model/model_reader.py
alec-deason: Ok

src/cascade/model/model_writer.py
adolgert: The user doesn't call this. The session does. And, yes, it sucks. Thank Brad. I didn't catch that this needed to be done until implementing the fifth user example from Brad, and something broke because it needed a fake age_id put in to account for the data age range. The point is to get the extra age from the **data**, not the model. So it can't happen when writing ages and times.

src/cascade/model/model_writer.py
adolgert: Sure. This is to make append make more sense.

src/cascade/model/model_writer.py
adolgert: I feel like hints say it's a list of dict, and I want to say it's a list of dicts of rates.

src/cascade/model/model_writer.py
alec-deason: Hmm. Ok. Wait does the model writer not also write the data?

src/cascade/model/model_writer.py
alec-deason: I guess. I don't really know what the right thing is. But when I see detailed type comments it feels like something more formal is called for. Maybe I'm wrong.

src/cascade/model/model_writer.py
alec-deason: Because data isn't part of the model? And this is just writing the model... Ok

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+        # further construction of grids.
+        prior_table = complete_table.sort_values(by="prior_id").reset_index(drop=True)[priors_columns]
+        if not self._dismod_file.prior.empty:
+            self._dismod_file.prior = self._dismod_file.prior.append(prior_table)
+        else:
+            self._dismod_file.prior = prior_table
+        return complete_table
+
+    def _flush_ages_times_locations(self):
+        if self._flushed:
```
alec-deason: So there's no way for the system to become unflushed? Once I've reached this point I can never add data to this writer. Is that guaranteed in any way?

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+    for check_ref_col in covariates:
+        if not isinstance(check_ref_col.reference, Real):
+            null_references.append(check_ref_col.name)
+    if null_references:
+        raise RuntimeError(f"Covariate columns without reference values {null_references}")
+
+    # Dismod requires us to rename covariates from names like sex, and "one"
+    # to x_0, x_1,... They must be "x_<digit>".
+    renames = dict()
+    for cov_idx, covariate in enumerate(covariates):
```
alec-deason: Sort? It would be nice if what we write is stable relative to minor perturbations. I haven't actually been watching for that. I'm actually not sure how important it really is. But it feels like something we'll miss eventually. Or maybe no human will ever have to look at one of these sqlite files ever again and that would be fine.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+            locations (pd.DataFrame): Initialize here because data refers to this.
+            parent_location (int): The session uses parent location to subset
+                                   data, but it isn't in the model.
+            filename (str|Path): Location of the Dismod db to overwrite.
+        """
+        self.dismod_file = DismodFile()
+        self._filename = Path(filename)
+        if self._filename.exists():
+            MATHLOG.info(f"{self._filename} exists so overwriting it.")
+            self._filename.unlink()
```
alec-deason: Good. I've been doing this by hand which is dumb.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+        self.dismod_file.engine = get_engine(self._filename)
+        self.parent_location = parent_location
+
+        self._create_node_table(locations)
+        self._basic_db_setup()
```
alec-deason: Even if _basic_db_setup doesn't do anything that _create_node_table depends on it is arresting to see the setup function called after the first function that touches the database.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+            initial_guess (Var): Starting point to look for solutions. If not
+                given, then the mean of the priors is taken as the initial
+                guess.
+
+        Returns:
+            DismodGroups[Var]: A set of fit var.
+        """
+        if model.random_effect:
+            MATHLOG.info(f"Running fit both.")
+            return self._fit("both", model, data, initial_guess)
```
alec-deason: Is there ever a case where I want to take a model that has random effects configured and only fit on the fixed effects anyway?

alec-deason: Oh, I see. There is a separate fit_fixed method.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+            model (Model): A model, possibly without scale vars.
+            data (pd.DataFrame): Data to fit.
+            initial_guess (Var): Starting point to look for solutions. If not
+                given, then the mean of the priors is taken as the initial
+                guess.
+
+        Returns:
+            DismodGroups[Var]: A set of fit var.
+        """
+        return self._fit("random", model, data, initial_guess)
```
alec-deason: Should this check that there are random effects in the model? Or do we just wait for dismod to compalin?

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+        return FitResult(self, self.get_var("fit"))
+
+    def _setup_model(self, model, data):
+        extremal = list()
+        if data is not None and not data.empty:
+            for dimension in ["age", "time"]:
+                cols = [ac for ac in data.columns if ac.startswith(dimension)]
+                if not cols:
+                    raise ValueError(f"Dataframe must have age and time columns but has {data.columns}.")
+                extremal.append({data[cols].min().min(), data[cols].max().max()})
```
alec-deason: The extreme ages and times might also appear in the priors, right?

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+        self.write_model(model, extremal)
+        self.write_data(data)
+        self._run_dismod(["init"])
```
alec-deason: It sometimes convenient to be able to setup the database without actually invoking dismod at all. For debugging. I'm not sure we want to support that and maybe it's less important now that things are more mature.

src/cascade/model/model_writer.py
```diff
@@ -0,0 +1,356 @@
...
+        for create_name in owned_tables:
+            if not getattr(self._dismod_file, create_name).empty:
+                cleared_tables.append(create_name)
+                setattr(self._dismod_file, create_name, self._dismod_file.empty_table(create_name))
+
+        # Handle rate separately.
+        if not self._dismod_file.rate.empty:
+            cleared_tables.append("rate")
+            # Then the rate has all five entries.
+            self._dismod_file.rate.loc[:, "parent_smooth_id"] = nan
```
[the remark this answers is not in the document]
adolgert: The empty tables method does create tables with the correct dtypes. Maybe they are already nan... I'll check.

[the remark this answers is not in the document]
adolgert: Nope. I think the point here isn't to make an empty table. There is initialization for the rates that happens whenever there is a new session. This method is meant to clear out the table from settings when there is a new model to write. The session is responsible for making a rate table that's emtpy, an integrand table, and a density table. Every time.

src/cascade/model/model_writer.py
adolgert: That's *exactly* what's happening.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+            ``max_difference`` from the ``reference`` covariate value.
+            Columns in the ``predicted`` are ``predict_id``, ``sample_index``,
+            ``avg_integrand`` (this is the value), ``location``, ``integrand``,
+            ``age_lower``, ``age_upper``, ``time_lower``, ``time_upper``.
+        """
+        self._check_vars(var)
+        model = Model.from_var(var, parent_location, weights=weights, covariates=covariates)
+        extremal = ({avgint.age_lower.min(), avgint.age_upper.max()},
+                    {avgint.time_lower.min(), avgint.time_upper.max()})
+        self.write_model(model, extremal)
```
alec-deason: This is not using the _setup_model method because that is only for fitting? Can we make that more general so it can be used in this case as well. Or make the name more specific so it's clear why it's not used here.

src/cascade/model/model_writer.py
adolgert: The writer defines a protocol that has to be called in order, start to finish. It's a rule. The writer is created, used, and deleted during the session's write_model.

src/cascade/model/model_writer.py
adolgert: They are sorted one level above this function. Always sex, then "one" covariate, and then the rest in sorted order. Very carefully. Greg still has his ways to debug the db files, and I'm trying to keep those working.

src/cascade/model/session.py
alec-deason: Ok, so it still works if there are extreme values which show up during the adding of age and time but don't appear in this initial list? Maybe we should call the list something like data_age_time_bounds or something to make that clear.

src/cascade/model/session.py
alec-deason: Yep, makes sense.

src/cascade/model/session.py
adolgert: Renamed to setup_model_for_fit. When fitting, we should be setting the start var and scale var. This does that. When doing predict, we should be setting the truth var, and that's what you see in this code. They are separate setups.

I renamed this to setup_model_for_fit.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+        self.flush()
+        with self._close_db_while_running():
+            str_command = [str(c) for c in command]
+            completed_process = run(["dmdismod", str(self._filename)] + str_command, stdout=PIPE, stderr=PIPE)
+            if completed_process.returncode != 0:
+                MATHLOG.error(completed_process.stdout.decode())
+                MATHLOG.error(completed_process.stderr.decode())
+            assert completed_process.returncode == 0, f"return code is {completed_process.returncode}"
+        if command[0] in COMMAND_IO:
+            self.dismod_file.refresh(COMMAND_IO[command[0]].output)
```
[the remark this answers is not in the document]
adolgert: Made possible by a very nice wrapper.

src/cascade/model/session.py
alec-deason: The current system has the --db-only flag. Though that may still run init. I can't remember. You could pass a similar flag here, though that's kind of ugly.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+        ))
+
+        # Defaults, empty, b/c Brad makes them empty even if there are none.
+        for create_name in ["nslist", "nslist_pair", "mulcov", "smooth_grid", "smooth"]:
+            setattr(self.dismod_file, create_name, self.dismod_file.empty_table(create_name))
+        self.dismod_file.log = make_log_table()
+        self._create_options_table()
+
+    def _create_options_table(self):
+        # Options in grey were rejected by Dismod-AT despite being in docs.
```
alec-deason: In grey? Commented out? But I don't see any...

src/cascade/model/model_writer.py
adolgert: The data isn't part of the model. That's totally it. It's a stats definition of what a model is.

src/cascade/model/model_writer.py
adolgert: This code is at a point that, yes, type hints could start to make sense. I see where you're coming from.

src/cascade/model/session.py
```diff
@@ -0,0 +1,552 @@
...
+            if np.isnan(location_id):
+                return np.nan
+            return np.where(table.c_location_id == location_id)[0][0]
+
+        table["parent"] = table.parent.apply(location_to_node_func)
+        self.dismod_file.node = table
+        self.location_func = location_to_node_func
+
+
+class FitResult:
```
alec-deason: I like this pattern of result objects.

src/cascade/model/session.py
adolgert: Renamed.

src/cascade/model/session.py
adolgert: I'd rather put an option on the session to a) keep a copy of the db before each run of a Dismod-AT command or b) never run Dismod-AT and mock all return values.

adolgert: Let's call that a new feature. Not on this PR.

src/cascade/model/session.py
adolgert: I figured out what the options really are by working through all examples. Thanks for catching this.

src/cascade/model/session.py
adolgert: Yes! Comes from R, to my knowledge. It lets you get data conditionally, but make that transparent to the user. There could be a problem if you hold onto the object and change the dismodfile. That isn't handled here.

src/cascade/model/smooth_grid.py
```diff
@@ -0,0 +1,153 @@
...
+        The Smooth Grid is a set of priors on an age-time grid.
+
+        Args:
+            age_time_grid (Tuple(set,set)): The supporting grid.
+        """
+        self.ages = np.sort(np.array(age_time_grid[0], dtype=np.float))
+        self.times = np.sort(np.array(age_time_grid[1], dtype=np.float))
+        self._view = dict()
+        for create_view in PriorKindEnum:
+            self._view[create_view.name] = _PriorView(create_view.name, self.ages, self.times)
```
alec-deason: Is view really the right term here? This isn't providing a view into another datastructure, it is the datastructure.

src/cascade/model/session.py
alec-deason: Yep

src/cascade/model/smooth_grid.py
adolgert: You're right. It was a view into a single dataframe, initially. Now it's not. Changed to `_PriorGrid`.

src/cascade/model/var.py
```diff
@@ -0,0 +1,109 @@
...
+        self.mulstd[kind].loc[:, "mean"] = value
+
+    def get_mulstd(self, kind):
+        return float(self.mulstd[kind]["mean"])
+
+    def __str__(self):
+        return f"Var({len(self.ages), len(self.times)})"
+
+    def __call__(self, age, time):
+        """Call a Var as a function of age and time.
```
alec-deason: I really like this interface.

adolgert: OMG, yes. It makes something that seemed hard, "bivariate interpolation given a Pandas DataFrame", and turn it into an automatic tool. Covers over complexity in ensuring that grids match each other, too.

tests/model/test_model.py
```diff
@@ -0,0 +1,235 @@
...
+        c_location_id=[1],
+    ))
+    parent_location = 1
+    db_file = Path("rftest.db")
+    session = Session(locations, parent_location, db_file)
+
+    # This peeks inside the session to test some of its underlying functionality
+    # without doing a fit.
+    session.write_model(basic_model, ([], []))
+    session._run_dismod(["init"])
```
alec-deason: Doesn't session.write_model already run init?

adolgert: No, it doesn't. We should be able to test `write_model` against an in-memory database, at some point. It does call flush() but not Dismod-AT.

alec-deason: Ok

PR #140, opened 2019-01-19
Clean up asyncio style

alec-deason:
Take a look. I think it's trivial and safe.

### DIFF
--- a/src/cascade/executor/dismod_runner.py
+++ b/src/cascade/executor/dismod_runner.py
@@ -131,8 +131,7 @@ def reduce_process_priority():
     os.nice(19)
 
 
-@asyncio.coroutine
-def _read_pipe(pipe, callback=lambda text: None):
+async def _read_pipe(pipe, callback=lambda text: None):
     """Read from a pipe until it closes.
 
     Args:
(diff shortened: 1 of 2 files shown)


src/cascade/executor/dismod_runner.py
```diff
@@ -188,8 +187,7 @@ def dismod_report_stderr(text):
         MATHLOG.warning(text, extra=dict(is_dismod_output=True))
 
 
-@asyncio.coroutine
-def async_run_and_watch(command, single_use_machine, poll_time):
+async def async_run_and_watch(command, single_use_machine, poll_time):
```
adolgert: Oh, you used the async keyword! Nice change.

src/cascade/executor/dismod_runner.py
```diff
@@ -211,10 +209,10 @@ def async_run_and_watch(command, single_use_machine, poll_time):
     std_out_task = loop.create_task(_read_pipe(sub_process.stdout, dismod_report_info))
     std_err_task = loop.create_task(_read_pipe(sub_process.stderr, dismod_report_stderr))
 
-    yield from sub_process.wait()
+    await sub_process.wait()
 
-    yield from std_out_task
-    yield from std_err_task
+    await std_out_task
+    await std_err_task
```
adolgert: This does seem more official.

alec-deason:
Yeah, there's nothing here. Just an update from the deprecated syntax to the modern syntax.

PR #142, opened 2019-01-24
documentation of the interactive interface

adolgert:
This documents how to use the new API. Wanted to get it online this morning for Greg.
Tried not to pollute the code, so lots is added through the Sphinx docs, explicitly.


docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+    data_cnt = 60
+    avgints = pd.DataFrame(dict(
+        integrand="susceptible",
+        location=parent_location,
+        age_lower=np.linspace(0, 120, data_cnt),
+        age_upper=np.linspace(0, 120, data_cnt),
+        time_lower=2000,
+        time_upper=2000,
+    ))
+    avgints = pd.concat([avgints, avgints.assign(integrand="mtother")])
```
alec-deason: This is concise but unclear. For a tutorial I would explicitly build out both the susceptible and mtother dataframes or try to come up with something equally obvious.

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+Those are the inputs for running Dismod-AT predict. The next step is to
+start a Dismod-AT session and run predict. A Dismod-AT session
+communicates with Dismod-AT running underneath.
+
+.. code:: python
+
+    session = Session(locations, parent_location, Path("fit0.db"))
+    session.set_option(ode_step_size=1)
+
+Note the ``ode_step_size=1``. This is the step size for predict. We can
```
alec-deason: "integration step size"? I don't know how explicit you need to be

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+That's the whole model. We could set the standard deviation on the
+age-difference priors from the forward age difference if we wanted, but
+it's enough to let them have a very large standard deviation.
+
+While that was 8 lines, the output of predict needs some help to look
+like input data.
+
+.. code:: python
+
+    data = predicted.drop(columns=["sample_index"]) \
```
alec-deason: I thought you were doing to do this normalization before you returned the results?

alec-deason: avg_integrand -> mean, is the normalization I mean. Not sample_index

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+        .rename(columns={"avg_integrand": "mean"})
+    data = data.assign(density="gaussian", std=0.3, eta=1e-4, nu=nan)
```
alec-deason: I think it's  worth explaining what this line does. Maybe it will be self evident to a modeler...

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+    
+    # It doesn't matter when the times are, because the system is constant
```
alec-deason: "system" -> "our model" or something. I think saying "system" makes it seem like this might somehow be a property of dismod which is confusing.

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+    # across time.
+    data.loc[:, "time_lower"] = np.random.randint(1990, 2015, size=len(data))
```
alec-deason: Wait, why random? Just to drive home the point that it doesn't matter? I think that's unnecessary.

alec-deason: Or to make it not constant across the data points... I guess that makes sense. Hmm. Still seems weird for an example like this.

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+    # We can add some range to ages and times if we want.
+    add_some_range_to_times = 0
+    add_size = add_some_range_to_times * np.random.random(size=len(data))
+    data.loc[:, "time_upper"] = data.loc[:, "time_lower"] + add_size
+    
+    # Want to intentionally take a data range and make it wrong?
+    mess_factor = 1  # 1 means leave it alone.
```
alec-deason: Cute, I like it.

docs/user/FitTutorial.rst
```diff
@@ -0,0 +1,279 @@
...
+        plt.show()
+
+.. code:: python
+
+    plot_mortality(mortality, result, data, value_residuals)
+
+
+.. parsed-literal::
+
+    /home/adolgert/.local/share/virtualenvs/cascade-f6I11cmg/lib/python3.6/site-packages/matplotlib/figure.py:2359: UserWarning: This figure includes Axes that are not compatible with tight_layout, so results might be incorrect.
```
alec-deason: You probably didn't mean to have this here...

docs/user/FitTutorial.rst
adolgert: The step size needs to become its own setting, not an option. And, yeah. I'll try.

docs/user/FitTutorial.rst
adolgert: Hah! Matplotlib is hard to wrangle.

PR #149, opened 2019-02-04
Bugfix/wrong var table ages

adolgert:
Big problem: avgint table is written incorrectly. And seems to be a problem with ages and times getting messed up, maybe with the vars.

Approach: Split out the part of the Session that deals with treating the sqlite3 file as objects, instead of dataframes. Then test from there. Don't create new functionality around better caching, as I wanted, because my team says to refactor and fix first.

NB: The ages and times tables, at one point, had to be sorted. That's no longer true, so that frees us up to get rid of "extremal_ages_and_times" part of the code, which was an annoying part.

1. This creates the ObjectWrapper class. Our dismod wrapper presents tables as dataframes. This presents them as objects. All code is refactored from the Session, which was too big anyway.
2. Tests the ObjectWrapper by using the python sqlite3 module to make sure it wrote what it said it wrote, for real.
3. **Writing a model deletes the current file and rewrites from scratch.** That's the big simplification.


src/cascade/model/object_wrapper.py
```diff
@@ -0,0 +1,315 @@
...
+
+    @property
+    def model(self):
+        raise NotImplementedError("Cannot read a model.")
+
+    @model.setter
+    def model(self, new_model):
+        """When you write a model, it deletes the file."""
+        self.make_new_dismod_file(self._locations_df)
+        writer = ModelWriter(self, {})
```
adolgert: Get rid of now-extra argument. (from Greg)

src/cascade/model/data_read_write.py
```diff
@@ -80,15 +108,24 @@ def avgint_to_dataframe(dismod_file, avgint, covariate_rename):
...
     _check_column_assigned(with_id, "integrand")
     with_weight = with_id.assign(weight_id=with_id.integrand.apply(lambda x: INTEGRAND_TO_WEIGHT[x].value))
     with_weight = with_weight.drop(columns=["integrand"]).reset_index(drop=True)
     with_location = with_weight.merge(
         dismod_file.node[["c_location_id", "node_id"]], left_on="location", right_on="c_location_id") \
         .drop(columns=["c_location_id", "location"])
+    _add_data_age_time_range(dismod_file, avgint)
     with_location = with_location.rename(columns=covariate_rename)
-    return with_location.assign(avgint_id=with_location.index)
+    return with_location.sort_values(by="avgint_id").reset_index(drop=True)
```
adolgert: This sort is unnecessary if you merge with how=left (Greg)

src/cascade/model/object_wrapper.py
```diff
@@ -0,0 +1,315 @@
...
+        new_model.write(writer)
+        writer.close()
```
adolgert: The business about making a new file... should flush every time you write anything. It's called isolation level for sqlite3. "exclusion" seems to be right. Take a look. (Greg again)

PR #150, opened 2019-02-05
comments on bundle upload

adolgert:
Documentation on how to do the data uploads we're doing now of full datasets.

### DIFF
--- a/docs/dev/bundle_upload.rst
+++ b/docs/dev/bundle_upload.rst
@@ -7,18 +7,6 @@ The goal is to upload a new synthetic dataset for the whole world.
 These synthetic datasets are constructed so that we have a known
 true answer (called an oracle) for figuring out how to fit.
 
-Most people are uploading a bunch of data and the maybe going back and
-adjusting a few rows for typos or whatever.
-But we're re-uploading whole data sets over and over. It's a different pattern.
-
-It's something we can automate. We just need to know what the ``seq`` are
-for the existing bundle. There's a function for that.
-Then you can either issue a deletion request separately or append the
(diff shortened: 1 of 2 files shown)


docs/dev/bundle_upload.rst
```diff
@@ -0,0 +1,48 @@
+.. _bundle-upload-full-datasets:
+
+Bundle Upload for Full Datasets
+===============================
+
+The goal is to upload a new synthetic dataset for the whole world.
+These synthetic datasets are constructed so that we have a known
+true answer (called an oracle) for figuring out how to fit.
+
+Most people are uploading a bunch of data and the maybe going back and
```
alec-deason: Move this paragraph and the one below to where the sentences following the one on line 48 are, and throw those sentences out since they are a shorter version of this.

docs/dev/bundle_upload.rst
```diff
@@ -0,0 +1,48 @@
...
+
+ 2. Then delete the existing bundle data using the modify bundle script
+    on the cluster::
+
+        /ihme/code/dismod_at/bin/modify_bundle --bundle-id 4319 \
+            /homes/alecwd/delete.xlsx
+
+    That assumes that your new bundles have the same number of
+    rows as the old ones. If they don't you'll need to make
+    a new deletion spreadsheet.
```
alec-deason: "To make a deletion spreadsheet use `elmo.run.get_bundle_data` to retrieve the existing bundle, save that to a spreadsheet, delete all columns other than `bundle_id` and `seq`. That gives you an upload sheet that when run through elmo will delete all rows from the bundle."

PR #151, opened 2019-02-06
Feature/resource tracking

alec-deason:
Integrates rocketsonde for resource usage tracking. See: https://github.com/ihmeuw/rocketsonde


src/cascade/dismod/db/metadata.py
```diff
@@ -335,6 +335,14 @@ class Option(Base):
     option_value = Column(String(), nullable=False)
 
 
+class Metadata(Base):
+    __tablename__ = "c_metadata"
+
+    c_metadata_id = Column(Integer(), primary_key=True, autoincrement=False)
+    key = Column(String(), unique=True)
+    value = Column(String(), nullable=False)
+
```
adolgert: This works. Metadata seems like it could refer to something dismod_at creates, or something about the model. Maybe c_execution_data, or something.

src/cascade/executor/epiviz_runner.py
```diff
@@ -362,22 +366,22 @@ def _check_dismod_command(dismod_file, command):
         raise DismodATException(f"DismodAt failed to complete '{command}' command")
 
 
-def run_dismod(dismod_file, command, *args):
-    dm_file_path = _get_dismod_db_path(dismod_file)
+def run_dismod(ec, command, *args):
+    dm_file_path = _get_dismod_db_path(ec.dismodfile)
```
adolgert: Yes, this is the opposite direction.

src/cascade/executor/epiviz_runner.py
```diff
@@ -523,33 +529,42 @@ def main(args):
     """
     start_time = default_timer()
     ec = make_execution_context()
+    ec.resource_monitor.start_monitor()
+    ec.resource_monitor.attach_to_process(os.getpid(), {"command": " ".join([sys.executable] + sys.argv)})
```
adolgert: Oh, you make one for this process. Nice.

src/cascade/executor/epiviz_runner.py
```diff
@@ -559,8 +574,8 @@ def one_location_set(ec, settings, posterior_draws_of_previous_fit):
     ec.dismodfile = write_dismod_file(mc, ec, ec.parameters.db_file_path)
     sampled_fit = None
     if not ec.parameters.db_only:
-        run_dismod(ec.dismodfile, "init")
-        run_dismod_fit(ec.dismodfile, has_random_effects(mc))
```
adolgert: Again, opposite direction.

adolgert: But maybe you want the context to take care of whether it's a file we run? Unsure.

src/cascade/executor/usage_reporter.py
```diff
@@ -0,0 +1,44 @@
...
+CREATE TABLE epi.dismod_resource_usage (
+    run_id varchar(32),
+    jobid varchar(255),
+    job_key text,
+    process_key text,
+    scale_parameters text,
+    inserted_at datetime,
+    summary text
+)
+"""
```
adolgert: What is this stray quotation up to?

src/cascade/executor/usage_reporter.py
```diff
@@ -0,0 +1,44 @@
...
+    with cursor(execution_context) as c:
+        c.execute(
+            q,
+            args={
+                "run_id": run_id.hex,
+                "jobid": jobid,
+                "job_key": c.connection.escape_string(job_key),
+                "process_key": c.connection.escape_string(process_key),
+                "scale_parameters": c.connection.escape_string(scale_parameters),
+                "summary": c.connection.escape_string(summary),
```
adolgert: The dumps and escape string seem in order. Can dumps ever throw an exception?

src/cascade/testing_utilities/__init__.py
```diff
@@ -9,4 +10,5 @@ def make_execution_context(**parameters):
     defaults.update(parameters)
     context = ExecutionContext()
     context.parameters = defaults
+    context.parameters.run_id = uuid4()
```
adolgert: This makes some sense for execution context.

tests/conftest.py
```diff
@@ -11,6 +13,7 @@ def mock_execution_context(mocker):
     mock_context.parameters.database = "test_database"
     mock_context.parameters.model_version_id = 12345
     mock_context.parameters.add_csmr_cause = 173
+    mock_context.parameters.run_id = UUID(bytes=b'1' * 16)
```
adolgert: Cute UUID.

src/cascade/dismod/db/metadata.py
alec-deason: Yeah, that makes sense.

src/cascade/executor/epiviz_runner.py
alec-deason: It does stuff. It's likely never the bottle neck but nice to have information about it.

src/cascade/executor/usage_reporter.py
alec-deason: Oh. Hmm. I put that there to keep track of my creation statement because we don't have an established way of tracking schema state. I don't really want to set that up for this one table but I feel weird not having a record of the schema somewhere. I don't know. I didn't mean to leave this orphaned like it is but I'm not sure where it belongs.

src/cascade/executor/usage_reporter.py
alec-deason: The dumps serialize to json. The escape string assures that the resulting string is safe to use in a SQL query, which it may not be as raw json.

src/cascade/executor/epiviz_runner.py
alec-deason: I was starting to juggle a lot of mysterious parameters for these functions so it made sense to me to pack some of them into the context object. But if we want to use the contexts less then that is the wrong motion. I don't feel strongly either way.

adolgert: This seems good here. I want to move to a point where I can guess which information is where. That needs to be addressed elsewhere, not this PR.

src/cascade/executor/usage_reporter.py
adolgert: Oh, right. When I have a system over which I have control, I make management functions or management command-line scripts. create_tables(), destroy_tables(), archive_tables(). Maybe put it into a create_tables() command?

src/cascade/executor/usage_reporter.py
```diff
@@ -0,0 +1,44 @@
...
+
+    jobid = None
+    if os.environ.get("JOB_ID"):
+        jobid = os.environ.get("JOB_ID")
+        if os.environ.get("UGE_TASK_ID"):
+            jobid += "." + os.environ.get("UGE_TASK_ID")
+    summary = json.dumps(summary, sort_keys=True)
+    job_key = json.dumps(job_key, sort_keys=True)
+    process_key = json.dumps(process_key, sort_keys=True)
+    scale_parameters = json.dumps(scale_parameters, sort_keys=True)
```
adolgert: What I mean about errors is that these json.dumps can throw TypeError or ValueError if something isn't the right type to encode in JSON. Should this have try-catch?

PR #153, opened 2019-02-07
make omega overlap data correctly

alec-deason:


src/cascade/executor/epiviz_runner.py
```diff
@@ -148,7 +148,11 @@ def add_omega_constraint(model_context, execution_context, sex_id):
         min_time = np.min(list(model_context.input_data.times))  # noqa: F841
         max_time = np.max(list(model_context.input_data.times))  # noqa: F841
         # The % 5 is to exclude annual data points.
-        asdr = asdr.query("time_lower >= @min_time and time_upper <= @max_time and time_lower % 5 == 0")
+        asdr = asdr.query(
+            "(time_lower >= @min_time and time_lower <= @max_time) or "
+            "(time_upper >= @min_time and time_upper <= @max_time) and "
+            "time_lower % 5 == 0"
+        )
```
adolgert: It seems to behave as expected when I make a test:
`import pandas as pd


def trial():
    point_times = pd.DataFrame(dict(
        time_lower=[1990, 2000, 2010], time_upper=[1990, 2000, 2010], values=[1,2,3]))
    interval_times = pd.DataFrame(dict(
        time_lower=[1990, 2000, 2010], time_upper=[1991, 2001, 2011], values=[1,2,3]))
    intervals = [(2000, 2000), (1995, 1995), (1970, 1980), (1995, 2005),
        (2005, 2015), (2010, 2010), (2015, 2017)]

    for times in [point_times, interval_times]:
        print(f"times {times}")

        for l, u in intervals:
            one_ans = times.query(
                "(time_lower >= @l and time_lower <= @u) or "
                "(time_upper >= @l and time_upper <= @u) and "
                "time_lower % 5 == 0"
                )
            print(f"{l} {u}\n{one_ans}\n")


if __name__ == "__main__":
    trial()
`
Upon further review, however, it doesn't pick up stuff near. What if we take the list of time points in the asdr data that are on the 5s, and then we include all those that are inside the min and max or surround it? So if the min/max is at 2003, you get 2000 and 2005. If the min/max is at 2000, you get 1995, 2000, 2005.

PR #157, opened 2019-02-09
Feature/dismodel cascade main

adolgert:
Took a little while b/c dependencies got squirrelly with old code.

 - Makes new Cascade Plan.
 - Makes an EstimationParameters class to describe what were ModelContext.


src/cascade/executor/dismodel_main.py
```diff
@@ -0,0 +1,103 @@
...
+
+def main(args):
+    start_time = default_timer()
+    execution_context = make_execution_context(gbd_round_id=5)
+    settings = load_settings(execution_context, args.meid, args.mvid, args.settings_file)
+    locations = location_hierarchy(execution_context)
+    plan = CascadePlan.from_epiviz_configuration(locations, settings)
+
+    for cascade_task_identifier in plan.cascade_jobs:
+        cascade_job, this_location_work = plan.cascade_job(cascade_task_identifier)
```
alec-deason: Is there a reason not to have `plan.cascade_jobs` produce these items directly, maybe with a task_identifier also, rather than requiring the second lookup?

src/cascade/executor/dismodel_main.py
```diff
@@ -0,0 +1,103 @@
...
+    parser.add_argument("-b", "--bundle-file", type=Path)
+    parser.add_argument("-s", "--bundle-study-covariates-file", type=Path)
+    parser.add_argument("--skip-cache", action="store_true")
+    parser.add_argument("--num-processes", type=int, default=4,
+                        help="How many subprocesses to start.")
+    parser.add_argument("--pdb", action="store_true")
+    args = parser.parse_args(args)
+
+    try:
+        software_version = get_distribution("cascade").version
```
alec-deason: Should we put this in `cascade.core.__version__`?

src/cascade/executor/cascade_plan.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        self._locations = None
+        self._task_graph = None
+        self._settings = settings
+
+    @property
+    def cascade_jobs(self):
+        return nx.lexicographical_topological_sort(self._task_graph)
+
+    def cascade_job(self, cascade_job_id):
```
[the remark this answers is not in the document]
adolgert: Yes, it's a tuple of (location_id, some index). I made that choice before it was being used, so that's premature design and a problem.

src/cascade/executor/cascade_plan.py
```diff
@@ -0,0 +1,121 @@
...
+        plan = cls(settings)
+        plan._locations = locations
+        if hasattr(settings.model, "drill_location_start") and \
+                settings.model.drill_location_start and settings.model.drill_location_end:
+            try:
+                drill = location_id_from_start_and_finish(
+                    plan._locations, settings.model.drill_location_start, settings.model.drill_location_end)
+            except ValueError as ve:
+                raise InputDataError(f"Location parameter is wrong in settings.") from ve
+        elif settings.model.drill_location:
```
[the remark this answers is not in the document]
adolgert: Yes, it is.

src/cascade/executor/cascade_plan.py
```diff
@@ -0,0 +1,121 @@
...
+        else:
+            MATHLOG.error(f"Looking for drill start and finish and cannot find "
+                          f"drill location start and end.")
+            raise InputDataError(f"Missing drill location start and end.")
+        MATHLOG.info(f"drill nodes {', '.join(str(d) for d in drill)}")
+        tasks = [(drill_location, 0) for drill_location in drill]
+        task_pairs = list(zip(tasks[:-1], tasks[1:]))
+        plan._task_graph = nx.DiGraph()
+        plan._task_graph.add_nodes_from(tasks)
+        plan._task_graph.add_edges_from(task_pairs)
```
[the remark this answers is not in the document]
adolgert: Yes and yes.

src/cascade/executor/dismodel_main.py
adolgert: Yes. Not only can the "this location work" be of considerable size, including locations for instance, but also having an identifier can make it possible to store in a file. I make an HDF or a zip file of dbs, and I identify each with its identifier.

src/cascade/executor/dismodel_main.py
adolgert: Why in core? I'm not opposed. I haven't seen that pattern. I could see putting it into `cascade.__version__` b/c that's a common pattern, like for numpy.

alec-deason: `cascade` is a namespace package, otherwise that is certainly where it would go

src/cascade/executor/dismodel_main.py
alec-deason: Cool

src/cascade/executor/cascade_plan.py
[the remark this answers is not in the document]
alec-deason: Ok. I'll ask Brian.

src/cascade/executor/cascade_plan.py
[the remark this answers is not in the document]
adolgert: I'm going to treat the identifier as opaque by using functions as accessors. That will make this something we can change.

src/cascade/executor/dismodel_main.py
adolgert: You want it in core b/c this is a namespace package. I figured out the answer, and I put it there.

PR #166, opened 2019-02-15
Feature/minimal running main

adolgert:
This is a working main.

- Constructs a model with rates and random effects but NO COVARIATES.
- Runs Dismod-AT for a fit.
- Does not hold out integrands.
- Does not constrain omega
- Does not make EMR from prevalence.
- Does not save results.
- Does not run fits using subprocesses in parallel.

Got it?


src/cascade/input_data/db/locations.py
```diff
@@ -6,13 +6,14 @@
...
 
-def location_hierarchy(execution_context):
+def location_hierarchy(gbd_round_id, location_set_version_id=None, location_set_id=None):
     """
     The GBD location hierarchy as a networkx graph where each node is the
     location id, and its properties are all properties returned by
     the dbtrees library. For instance
 
-    >>> locations = location_hierarchy(ec)
+    >>> locations = location_hierarchy(6, location_set_id=35)
```
alec-deason: Why mention both location_set_ids?

src/cascade/model/model.py
```diff
@@ -18,8 +18,8 @@ class Model(DismodGroups):
     """
     def __init__(self, nonzero_rates, parent_location, child_location=None, covariates=None, weights=None):
         """
-        >>> locations = location_hierarchy(execution_context)
-        >>> m = Model(["chi", "omega", "iota"], 6, locations)
+        >>> locations = location_hierarchy(6, age_group_set_id=35)
```
alec-deason: age_group_set_id as a parameter to location_hierarchy?

tests/executor/test_cascade_plan.py
```diff
@@ -12,62 +12,68 @@
 def test_create_start_finish(ihme):
     args = parse_arguments(["z.db"])
     ec = make_execution_context(parent_location_id=0, gbd_round_id=5)
-    locations = location_hierarchy(ec)
+    locations = location_hierarchy(6, location_set_id=35)
```
alec-deason: this is the wrong version set?

tests/executor/test_cascade_plan.py
```diff
@@ -12,62 +12,68 @@
...
-    assert len(c._task_graph.nodes) == 3
+    assert len(c._task_graph.nodes) == 4
     print(nx.to_edgelist(c._task_graph))
 
 
 def test_single_start_finish(ihme):
     args = parse_arguments(["z.db"])
     ec = make_execution_context(parent_location_id=0, gbd_round_id=5)
-    locations = location_hierarchy(ec)
+    locations = location_hierarchy(6, location_set_id=35)
```
alec-deason: wrong version set

tests/executor/test_cascade_plan.py
```diff
@@ -12,62 +12,68 @@
...
-    assert len(c._task_graph.nodes) == 1
+    assert len(c._task_graph.nodes) == 2
     print(nx.to_edgelist(c._task_graph))
 
 
 def test_iterate_tasks(ihme):
     args = parse_arguments(["z.db"])
     ec = make_execution_context(parent_location_id=0, gbd_round_id=5)
-    locations = location_hierarchy(ec)
+    locations = location_hierarchy(6, location_set_id=35)
```
alec-deason: same

tests/executor/test_priors_from_draws.py
```diff
@@ -0,0 +1,120 @@
...
+            draws = list()
+            for draw_idx in range(draw_cnt):
+                var = model.var_from_mean()
+                jitter(var, rng)
+                draws.append(var)
+            parent_model_has_random_effects = len(model.random_effect) > 0
+
+            set_priors_from_draws(model, draws)
+
+            seen = list()
```
alec-deason: This does seem to work but I had to think about how it would interact with the closure captured by _gather. I think I'd actually prefer `seen.clear()` then there's no question.

src/cascade/executor/estimate_location.py
```diff
@@ -55,17 +68,111 @@ def retrieve_data(execution_context, local_settings):
...
 
+    Returns:
+        The fit.
+    """
 
-def save_outputs(computed_fit, execution_context, local_settings):
+    session = Session(
+        locations=input_data.locations_df,
+        parent_location=model.location_id,
+        filename="subsession.db"
```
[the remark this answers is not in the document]
adolgert: This is so very not safe. Can you address this during running under asyncio? The execution context should know how / where to save any files.

src/cascade/executor/epiviz_runner.py
```diff
@@ -170,7 +170,7 @@ def add_omega_constraint(model_context, execution_context, sex_id):
     model_context.rates.omega.parent_smooth = build_constraint(parent_asdr)
     MATHLOG.debug(f"Add {parent_asdr.shape[0]} omega constraints from age-standardized death rate data to the parent.")
 
-    locations = location_hierarchy(execution_context)
+    locations = location_hierarchy(execution_context.parameters.gbd_round_id, location_set_id=35)
```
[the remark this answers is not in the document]
adolgert: changed to 429

src/cascade/executor/estimate_location.py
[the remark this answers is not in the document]
alec-deason: Is there a reason to not have the session stuff things in a tempdir unless you specify otherwise? And then you'd just leave the `filename` parameter off unless you wanted it someplace permanent for debugging or what not. Actually that answers my later question about what "closing" a session means. It means cleaning up the temporary directory.

PR #168, opened 2019-02-21
Bugfix/meas std effect

alec-deason:


src/cascade/model/object_wrapper.py
```diff
@@ -288,7 +288,7 @@ def _create_options_table(self):
         option = pd.DataFrame([
             dict(option_name="parent_node_id", option_value=str(self.location_func(self.parent_location))),
             dict(option_name="parent_node_name", option_value=nan),
-            dict(option_name="meas_std_effect", option_value="add_std_scale_all"),
+            dict(option_name="meas_std_effect", option_value="add_var_scale_log"),
```
adolgert: Huh. They want this one? OK.

alec-deason: That's what Greg said

PR #170, opened 2019-02-24
Feature/hdf5 serialize

adolgert:
Saves Model and Var to HDF5. I can extend the documentation after the documentation PR is merged.
This is in a separate module because it depends on the Model but the Executor depends on this.


src/cascade/serialize/hdf.py
```diff
@@ -0,0 +1,286 @@
...
+        datasets = [ds_name for ds_name in hdf_group.keys() if ds_name.startswith(group_name)]
+        for ds_name in datasets:
+            ds = hdf_group[ds_name]
+            if "dismod_group" in ds.attrs:
+                key0 = ds.attrs["key0"]
+                if "key1" in ds.attrs:
+                    try:
+                        key = (str(key0), int(ds.attrs["key1"]))
+                    except ValueError:
+                        key = (str(key0), str(ds.attrs["key1"]))
```
alec-deason: This can't be typed in the underlying format?

src/cascade/serialize/hdf.py
```diff
@@ -0,0 +1,286 @@
...
+    values datasets."""
+    ages = var.ages
+    times = var.times
+
+    ds = hdf_group.create_dataset(name, (len(ages), len(times)), dtype=np.float)
+    for aidx, tidx in product(range(len(ages)), range(len(times))):
+        ds[aidx, tidx] = var[ages[aidx], times[tidx]]
+
+    for scale_idx, kind in enumerate(["age", "time"]):
+        dimension = write_dimension(hdf_group, kind, [ages, times][scale_idx])
```
alec-deason: `[ages, times][scale_idx]` This is super weird. I actually think I'd prefer `enumerate(zip([ages, times], ["age", "time"]))` ore something. Though I think you can make a pretty strong argument that that is also super weird.

src/cascade/serialize/hdf.py
```diff
@@ -0,0 +1,286 @@
...
+
+    ds.attrs["cascade_type"] = "Var"
+    return ds
+
+
+def read_var(ds):
+    if "cascade_type" not in ds.attrs or ds.attrs["cascade_type"] != "Var":
+        raise SerializationError(f"Expected {ds} to be a var")
+
+    ages = ds.dims[0][0]
```
alec-deason: Maybe a constant AGE_DIM instead of the magic `0`?

docs/dev/hdf_design.rst
```diff
@@ -0,0 +1,93 @@
...
+
+The main tool is the library `h5py <http://docs.h5py.org/en/stable/index.html>`_,
+whose documentation is good. For more background, read
+the `HDF5 Documentatiion <https://support.hdfgroup.org/HDF5/doc/index.html>`_.
+
+This is a file format both for single runs of Dismod-AT through the Cascade
+and for full global runs. It have an atomic store-this-one-fit and a larger
+set of rules around how to store multiple fits for a global Cascade.
+
+For the global Cascade, separate observation from estimation from sampling
```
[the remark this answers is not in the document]
adolgert: Rewrote the paragraph, maybe just as confusingly, but trying. 1. Separate top-level groups help data management. 2. On a technical level, group structure doesn't affect where or how data is saved in the file's block structure, so don't worry about that part. 3. Links among objects make this all easier.

docs/dev/inputs_outputs.rst
```diff
@@ -0,0 +1,131 @@
...
+  inputs and settings, although they are difficult to parse. The reason
+  to parse them is to see how interactions between what the user requested
+  through the EpiViz-AT user interface and what the databases provide resulted
+  in a final model. There is one log per estimation run.
+
+Dismod-AT Logs
+  These get embedded in the Cascade logs, but it is worth mentioning that
+  the log-level used for Dismod-AT can make the logs get very large,
+  because Dismod-AT can log every step of its nonlinear solver. Dismod-AT
+  logs have two levels, and we really need the top-level log in order
```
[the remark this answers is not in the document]
adolgert: Dismod-AT's log messages versus Ipopt's log messages. Thanks for noticing b/c was confusing.

docs/dev/inputs_outputs.rst
```diff
@@ -0,0 +1,131 @@
...
+
+Fits
+  The output of the main fit and of the draws, which determine uncertainty,
+  is an interpolation over age and time. This is different from the draws
+  described below. It's specified by rate values on an age and time grid.
+  If there are seven years and 23 age points, there are 161 values.
+  Fits also include "standard deviation multipliers" which are hyper-priors
+  on the standard deviation of the priors. There are up to three per
+  underlying rate in the problem. There can be five rates, and each rate
+  generally has about 30 age points. A rate can have five year points or
```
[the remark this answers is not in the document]
adolgert: 1990, 1995, 2000, 2005, 2010, 2015, 2018... 7, which is about 5? This is very ballpark-y. Clarified.

src/cascade/serialize/hdf.py
adolgert: These are those )^@#%$#$ numpy array scalars. They are typed, but this is the only way I know to extract them.

src/cascade/serialize/hdf.py
adolgert: Went with a namedtuple with two members: name and coord. Looking better, but it does push it towards index 0 and index 1 instead of age and time.

PR #171, opened 2019-02-25
Feature/check json features

adolgert:
The main user feature this PR delivers is that I'm 100% sure we obey every part of the EpiViz-AT settings that goes into the Dismod-AT options table.

The mechanism underneath that makes that verification possible is the main code feature. It will enable us to test all the parts of the EpiViz-AT settings document to see that they do affect db creation as expected. This PR applies that to the options, but it can be applied to all parts of the settings.

In the code to create random settings, it is now possible to create a specific version of the settings. The idea is to make a version of the settings and then change little bits and watch how they change the data and model written to the db file.

### DIFF
--- a/src/cascade/executor/construct_model.py
+++ b/src/cascade/executor/construct_model.py
@@ -99,6 +99,9 @@ def construct_grid_ages_times(default_age_time, single_age_time, smooth):
 
 
 def construct_model_random_effects(default_age_time, single_age_time, ev_settings, model):
+    """The settings may have random effects for many locations which aren't children
+    of the current parent location. Only those random effects that apply to the children
+    explicitly or to all locations (specified as location=None) are included."""
     if not hasattr(ev_settings, "random_effect"):
         return
     if not ev_settings.random_effect:
@@ -114,7 +117,7 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
             else:
                 pass  # An unset prior should be unused (dage for one age, dtime for one time)
 
-        if hasattr(smooth, "location") and smooth.location is not None and smooth.location != model.location_id:
+        if hasattr(smooth, "location") and smooth.location in model.child_location:
             location = smooth.location
         else:
             # One smooth for all children when there isn't a child location.
--- a/src/cascade/executor/construct_model.py
+++ b/src/cascade/executor/construct_model.py
@@ -79,7 +79,7 @@ def construct_model_rates(default_age_time, single_age_time, ev_settings, model)
 
 
 def construct_grid_ages_times(default_age_time, single_age_time, smooth):
-    if hasattr(smooth, "age_time_specific") and smooth.age_time_specific == 0:
+    if not smooth.is_field_unset("age_time_specific") and smooth.age_time_specific == 0:
         return single_age_time
 
     ages = smooth.age_grid
@@ -102,8 +102,6 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
     """The settings may have random effects for many locations which aren't children
     of the current parent location. Only those random effects that apply to the children
     explicitly or to all locations (specified as location=None) are included."""
-    if not hasattr(ev_settings, "random_effect"):
-        return
     if not ev_settings.random_effect:
         return
 
@@ -117,7 +115,7 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
             else:
                 pass  # An unset prior should be unused (dage for one age, dtime for one time)
 
-        if hasattr(smooth, "location") and smooth.location in model.child_location:
+        if not smooth.is_field_unset("location") and smooth.location in model.child_location:
             location = smooth.location
         else:
             # One smooth for all children when there isn't a child location.
--- a/src/cascade/executor/create_settings.py
+++ b/src/cascade/executor/create_settings.py
@@ -12,8 +12,10 @@ grid at 1 point or >1 point for age or time
 priors present or absent on value, dage, dtime
 bounds upper/lower can differ between rates, res, and covariates.
 """
+from configparser import ConfigParser
 from copy import deepcopy
 from pprint import pprint
+from textwrap import dedent
 
 import networkx as nx
 from numpy.random import RandomState
@@ -95,7 +97,7 @@ BASE_CASE = {
 }
 
 
-def grid(non_negative, rng, single_age=False):
+def grid(non_negative, rng, name, single_age=False):
     grid_case = {
         "default": {
             "value": {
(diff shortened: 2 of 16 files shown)


src/cascade/executor/create_settings.py
```diff
@@ -211,45 +258,100 @@ def create_settings(rng, locations):
...
+    although this adds the section title for you.
+    """
+    def __init__(self, rng=None, settings=None):
+        if isinstance(settings, str):
+            parser = ConfigParser()
+            parser.read_string("[settings]\n" + dedent(settings))
+            self.answers = parser["settings"]
+        else:
+            self.answers = dict()
+        self.rng = rng
```
alec-deason: What would it mean for the rng to be None? It seems to me that if the user doesn't pass in their own randomstate we should make one for them.

src/cascade/executor/create_settings.py
```diff
@@ -211,45 +258,100 @@ def create_settings(rng, locations):
...
+                try:
+                    value = float(value)
+                except ValueError:
+                    if value.lower() == "true":
+                        return True
+                    elif value.lower() == "false":
+                        return False
+            return value
+        else:
+            assert self.rng, f"Unexpected setting chosen {name} {choices}"
```
alec-deason: This assert and message combination doesn't make sense to me. This is checking if the rng is None but the message has nothing to do with that. Oh, wait. I see. This is saying that if someone doesn't supply an RNG then they can only select from the preset values. Why support that use case?

src/cascade/executor/session_options.py
```diff
@@ -23,15 +23,17 @@ def make_options(ev_settings):
     options["ode_step_size"] = ev_settings.model.ode_step_size
     if ev_settings.model.additional_ode_steps:
         options["age_avg_split"] = " ".join(str(ea) for ea in ev_settings.model.additional_ode_steps)
-    if hasattr(ev_settings, "random_seed") and ev_settings.random_seed:
-        options["random_seed"] = ev_settings.random_seed
+    if hasattr(ev_settings.model, "random_seed") and ev_settings.model.random_seed:
```
alec-deason: If this field is in the form configuration it should always be present on the settings objects, even if it wasn't set by the users. If you are checking for whether the user supplied it then use `ev_settings.model.is_field_unset("random_seed")`

Also, this check does the wrong thing if the user specifies a random_seed of `0` (which they shouldn't but you know, technically....)

src/cascade/testing_utilities/compare_dismod_db.py
```diff
@@ -0,0 +1,45 @@
...
+        self.tables = [sorted([x[0] for x in connl.execute(SQLITE_TABLES)]) for connl in self.conn]
+
+    def table_diffs(self):
+        sm = SequenceMatcher()
+        sm.set_seqs(*self.tables)
+        return [op for op in sm.get_opcodes() if op[0] != "equal"]
+
+    def different_tables(self):
+        """Which tables are different, except for the log, which always differs."""
+        common = sorted(set(self.tables[0]) & set(self.tables[1]))
```
alec-deason: Are tables which are absent in one database not considered different?

src/cascade/executor/create_settings.py
adolgert: I want to be able to make a specific JSON settings with this tool and know that I'm getting the same one every time. If I failed to specify part of it, then it will throw this error, and I'll see what I failed to specify. The use case is using this system to make one set of settings.

src/cascade/testing_utilities/compare_dismod_db.py
adolgert: The `table_diffs` tells you which tables are added or deleted. This one tells you which of existing tables have differences that you can probe into. It makes more sense in the way it's used.

src/cascade/executor/session_options.py
adolgert: Alec, how does this work for something like zero_sum_random, that's a list? I don't know how to use is_field_unset() for this because lists don't have that method. If someone sets zero sum random to something and then wants to unset it... that's not a use case here, is it?

adolgert: Oh, I'm doing it wrong. It's on the form.

src/cascade/executor/construct_model.py
```diff
@@ -99,6 +99,9 @@ def construct_grid_ages_times(default_age_time, single_age_time, smooth):
 
 
 def construct_model_random_effects(default_age_time, single_age_time, ev_settings, model):
+    """The settings may have random effects for many locations which aren't children
+    of the current parent location. Only those random effects that apply to the children
+    explicitly or to all locations (specified as location=None) are included."""
     if not hasattr(ev_settings, "random_effect"):
```
alec-deason: hasattr

src/cascade/executor/construct_model.py
```diff
@@ -114,7 +117,7 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
             else:
                 pass  # An unset prior should be unused (dage for one age, dtime for one time)
 
-        if hasattr(smooth, "location") and smooth.location is not None and smooth.location != model.location_id:
+        if hasattr(smooth, "location") and smooth.location in model.child_location:
```
alec-deason: hasattr

src/cascade/executor/estimate_location.py
```diff
@@ -82,34 +82,31 @@ def modify_input_data(input_data, local_settings):
     ev_settings = local_settings.settings
     # These are suitable for input to the fit.
     if hasattr(ev_settings.eta, "data") and ev_settings.eta.data:
```
alec-deason: hasattr

src/cascade/executor/estimate_location.py
```diff
@@ -82,34 +82,31 @@ def modify_input_data(input_data, local_settings):
...
-        data_eta = float(ev_settings.eta.data)
+        data_eta = defaultdict(lambda: float(ev_settings.eta.data))
     else:
-        data_eta = None
+        data_eta = defaultdict(lambda: nan)
+    id_to_integrand = make_integrand_map()
+    for set_eta in ev_settings.data_eta_by_integrand:
+        data_eta[id_to_integrand[set_eta.integrand_measure_id]] = float(set_eta.value)
+
+    if hasattr(ev_settings.model, "data_density") and ev_settings.model.data_density:
```
alec-deason: hasattr

src/cascade/executor/session_options.py
```diff
@@ -11,39 +11,38 @@ def make_options(ev_settings):
...
         options["age_avg_split"] = " ".join(str(ea) for ea in ev_settings.model.additional_ode_steps)
-    if hasattr(ev_settings, "random_seed") and ev_settings.random_seed:
-        options["random_seed"] = ev_settings.random_seed
+    if not ev_settings.model.is_field_unset("random_seed"):
+        options["random_seed"] = ev_settings.model.random_seed
     # rate case is set for us.
 
     for kind in ["fixed", "random"]:
         for opt in ["derivative_test", "max_num_iter", "print_level", "accept_after_max_steps", "tolerance"]:
             if hasattr(ev_settings, opt):
```
alec-deason: Another hasattr, though this one may actually be necessary if the fixed and random forms have different fields

PR #172, opened 2019-02-26
write dismod output to logs in real time

alec-deason:


src/cascade/model/session.py
```diff
@@ -349,3 +345,62 @@ def simulation(self, index):
             the Nth simulation.
         """
         return self._file_objects.read_simulation_model_and_data(self._model, self._data, index)
+
```
adolgert: These functions are of a piece, and they might make sense in their own file?

src/cascade/model/session.py
```diff
@@ -349,3 +345,62 @@ def simulation(self, index):
...
+        std_out_task = loop.create_task(_read_pipe(sub_process.stdout, dismod_report_info(stdout_accumulator)))
+        std_err_task = loop.create_task(_read_pipe(sub_process.stderr, dismod_report_stderr(stderr_accumulator)))
+
+        await sub_process.wait()
+        await std_out_task
+        await std_err_task
+
+        return sub_process.returncode, "".join(stdout_accumulator), "".join(stderr_accumulator)
+
+    return loop.run_until_complete(coroutine())
```
adolgert: That all looked very familiar and has been working.

tests/model/test_session.py
```diff
@@ -42,3 +44,27 @@ def test_options(dismod):
     ))
     # Run a predict in order to verify the options are accepted.
     session.predict(model_var, avgints, parent_location)
+
```
adolgert: I like the bash test and /bin/false. I don't know what will happen when the run command fails b/c it isn't found. There's no error-condition testing. This is how it was, so I'm just commenting.

alec-deason: It raises a FileNotFound exception. We could wrap it in a custom exception but it's really not a recoverable error the way this tool gets used so I think it's a fine choice to just let it bubble out.

src/cascade/model/session.py
alec-deason: Bleh. I'm struggling with which package that file would be in. It clearly doesn't belong in model. It seems like it belongs in executor but I think you said you didn't want other packages to have dependencies on executor, which is totally reasonable. Ok I'm going to make `cascade.core.subprocess_utils` That's not perfect but I think it's good enough and if we want to move it later, that's easy.

PR #173, opened 2019-02-27
Upload fit data for children rather than parent

alec-deason:
We've talked this through and I think it's safe so I'm going to just go ahead and merge it so I can deploy but if you see issues, let me know.

### DIFF
--- a/src/cascade/model/integrands.py
+++ b/src/cascade/model/integrands.py
@@ -1,6 +1,7 @@
 import pandas as pd
 
 from cascade.input_data.db.demographics import get_age_groups, get_years
+from cascade.input_data.db.locations import get_descendants
 from cascade.dismod.constants import IntegrandEnum
 
 from cascade.core.log import getLoggers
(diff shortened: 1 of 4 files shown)


tests/conftest.py
```diff
@@ -17,6 +19,16 @@ def mock_execution_context(mocker):
     return mock_context
 
 
+@pytest.fixture
+def mock_locations(mocker):
+    locations = mocker.patch("cascade.input_data.db.locations.location_hierarchy")
+    G = nx.DiGraph()
+    G.add_nodes_from(list(range(8)))
+    G.add_edges_from([(0, 1), (0, 2), (1, 3), (1, 4), (2, 5), (2, 6), (6, 7)])
```
adolgert: nice.

src/cascade/model/integrands.py
```diff
@@ -20,6 +21,7 @@ def make_average_integrand_cases_from_gbd(execution_context, sexes, include_birt
     gbd_age_groups = get_age_groups(execution_context)
     age_ranges = [(r.age_group_years_start, r.age_group_years_end) for _, r in gbd_age_groups.iterrows()]
     time_ranges = [(y, y) for y in get_years(execution_context)]
+    locations = get_descendants(execution_context, children_only=True)
```
adolgert: This looks right.

tests/conftest.py
alec-deason: Pretty sure that's your mock. I just moved it.

PR #176, opened 2019-02-28
Fix some small bugs around sex_id and the retrieve_data function

alec-deason:


alec-deason:
Nope, that's perfect.

PR #177, opened 2019-03-02
Feature/drill covariate data country

adolgert:
The main feature is adding the ability to use covariate multipliers that use country covariates. This moves our covariate data retrieval and interpolation of covariate data to the new code base.
Half of this PR is asserting data types for the raw inputs.

Added functionality to add CSMR to the bundle. Taken from the old code. There is one main difference: the old code used a bundle field called "standard error" to estimate what Dismod-AT calls standard deviation. I think this was a mistake and is responsible for the too-small uncertainties seen in EpiViz-AT. I changed it to take uncertainty from the lower and upper bounds on the confidence interval, and I excised use of standard error for CSMR and the bundle.

And it adds the omega constraint.

And it filters the "excluded measures" from the bundle.


src/cascade/executor/estimate_location.py
```diff
@@ -122,23 +154,58 @@ def modify_input_data(input_data, local_settings, covariate_data_spec):
     for set_density in ev_settings.data_density_by_integrand:
         density[id_to_integrand[set_density.integrand_measure_id]] = set_density.value
 
+    csmr = normalize_csmr(input_data.cause_specific_mortality_rate, local_settings.sex_id)
+    CODELOG.debug(f"bundle cols {input_data.bundle.columns}\ncsmr cols {csmr.columns}")
+    assert not set(csmr.columns.tolist()) - set(input_data.bundle.columns.tolist())
```
alec-deason: You can convert `.columns` directly to a set. No need for tolist

src/cascade/executor/estimate_location.py
```diff
@@ -122,23 +154,58 @@ def modify_input_data(input_data, local_settings, covariate_data_spec):
...
         density,
+        nu,
     )
-    input_data.observations = input_data.observations.drop(columns="sex_id")
     # ev_settings.data_eta_by_integrand is a dummy in form.py.
     MATHLOG.info(f"Ignoring data_eta_by_integrand")
 
     input_data.locations_df = location_hierarchy_to_dataframe(input_data.locations)
+    # Add mortality data.
+    # Add EMR from prevalence.
```
alec-deason: Maybe make these TODOs

src/cascade/input_data/configuration/construct_bundle.py
```diff
@@ -39,6 +40,41 @@ def _normalize_measures(data):
...
+    else:
+        measures_to_exclude = list()
+    if ev_settings.policies.exclude_relative_risk:
+        measures_to_exclude.append("relrisk")
+    # else don't add relrisk to excluded measures
+
+    if measures_to_exclude:
+        mask = bundle.measure.isin(measures_to_exclude)
+        if mask.sum() > 0:
+            bundle = bundle[~mask]
```
alec-deason: You may not want this in this PR but I have a task in my notebook around this which I keep forgetting to act on. The modelers would like to be able to see which measures were excluded when they look at the data in epiviz. That would mean marking those rows as holdout when uploading to tier3

src/cascade/input_data/configuration/raw_input.py
```diff
@@ -0,0 +1,144 @@
+import numpy as np
+import pandas as pd
+from numpy import dtype
+
+from cascade.core import getLoggers
+
+CODELOG, MATHLOG = getLoggers(__name__)
+
+
+EXPECTED_TYPES = dict(
```
alec-deason: Good

src/cascade/testing_utilities/dtypes.py
```diff
@@ -0,0 +1,20 @@
...
+    >>> print_types(input_data)
+
+    """
+    with Path(f"dtypes{np.random.randint(0, 999)}.txt").open("w") as outfile:
+        for member_name in dir(data):
+            member = getattr(data, member_name)
+            if isinstance(member, pd.DataFrame):
+                dtypes = member.dtypes
+                as_data = list(zip(dtypes.index.tolist(), dtypes.tolist()))
+                print(f"data_type {member_name}: {as_data}", file=outfile)
```
alec-deason: Should this be a print or a log?

docs/dev/versioning.rst
```diff
@@ -3,8 +3,11 @@
 Versioning
 ==========
 
-We are using the date to version. This means tagging versions with git, and then
-the rest should work. The command is::
+We are using the date to version and ``setuptools_scm`` to create the version.
+This means tagging versions with git, and then
+the rest should work. A new version is created with each install and tagged
+by date, as enforced by the instal script. Were you to do it by hand,
```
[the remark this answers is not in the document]
adolgert: I can't figure out why I'm blind to the little underlines that show when spelling is wrong.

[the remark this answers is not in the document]
alec-deason: My stuff is riddled with this kind of thing so I certainly can't judge.

src/cascade/executor/cascade_plan.py
```diff
@@ -89,6 +89,12 @@ def cascade_job(self, cascade_job_id):
             grandparent_location_id = None
 
         parent_location_id = self._location_of_cascade_job(cascade_job_id)
+        if self._settings.model.is_field_unset("drill_sex"):
+            # An unset drill sex gets all data.
+            sex_id = [1, 2, 3]
+        else:
+            # Setting to male or female pulls in "both."
+            sex_id = [self._settings.model.drill_sex, 3]
```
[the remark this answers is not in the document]
adolgert: Let's enforce 4->3 at our boundaries. So I'd like to leave 3 here and add 4 during bundle filtering.

src/cascade/executor/construct_model.py
```diff
@@ -71,6 +91,53 @@ def construct_model_rates(default_age_time, single_age_time, ev_settings, model)
...
+
+        def child_effect(age, time):
+            return np.log(child_rate(age, time) / omega(age, time))
+
+        model.rate[("omega", child)] = constraint_from_rectangular_data(child_effect, default_age_time)
+
+
+def constraint_from_rectangular_data(omega, default_age_time):
+    """Takes data on a complete set of ages and times, makes a constraint grid."""
+    omega_grid = SmoothGrid(ages=default_age_time["age"], times=default_age_time["time"])
```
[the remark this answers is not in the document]
adolgert: I renamed it to `rate_var`, so at least we know now it's a var.

src/cascade/executor/covariate_data.py
```diff
@@ -49,17 +54,45 @@ def add_covariate_data_to_observations_and_avgints(data, local_settings, epiviz_
...
+    add_study_covariate_to_observations_and_avgints(data)
+    add_country_covariate_to_observations_and_avgints(data, local_settings, epiviz_covariates)
+
+
+def add_country_covariate_to_observations_and_avgints(data, local_settings, epiviz_covariates):
+    """Adds country covariates to observation and average integrand cases."""
+    country_specs = {ccov for ccov in epiviz_covariates if ccov.study_country == "country"}
+    for covariate_id in {evc.covariate_id for evc in country_specs}:
+        ccov_ranges_df = data.country_covariates[covariate_id]
+        print(f"Columns of ccof_df {ccov_ranges_df.columns}")
```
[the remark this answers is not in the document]
adolgert: dammit. I should _never_ put a print in there. A LOGGER.debug will always do.

src/cascade/executor/estimate_location.py
adolgert: or rather, they are done.

src/cascade/input_data/configuration/construct_bundle.py
adolgert: totally doable. Let me put that in and see how it goes.
`
        if mask.sum() > 0:
            bundle.loc[mask, "hold_out"] = 1
            MATHLOG.info(
`

src/cascade/testing_utilities/dtypes.py
adolgert: erm. That's under testing. I like the print there. Prints are bad, though... I'll change it.

adolgert: Wait. That's going to a file, on purpose. That function was there to snoop on what comes in. I used it this way: I'd get stuff from the IHME db, not know what the columns are, and use this to see if they varied... You know, I should do that with the debugger... aaaaand deleted the file.

alec-deason: Oh, I didn't even notice it was going to a file. But moot point I guess.

PR #179, opened 2019-03-05
Feature/random effect nslist for children

adolgert:
fixes checker

### DIFF
--- a/src/cascade/executor/construct_model.py
+++ b/src/cascade/executor/construct_model.py
@@ -1,3 +1,5 @@
+from collections import defaultdict
+
 import numpy as np
 
 from cascade.core.log import getLoggers
@@ -74,7 +76,9 @@ def construct_model(data, local_settings, covariate_multipliers, covariate_data_
     )
 
     construct_model_rates(default_age_time, single_age_time, ev_settings, model)
-    construct_model_random_effects(default_age_time, single_age_time, ev_settings, model)
+    # No random effects if there is only one child.
+    if children and len(children) > 1:
+        construct_model_random_effects(default_age_time, single_age_time, ev_settings, model)
     construct_model_covariates(default_age_time, single_age_time, covariate_multipliers, model)
     if ev_settings.model.constrain_omega:
         constrain_omega(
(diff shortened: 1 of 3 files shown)


src/cascade/executor/construct_model.py
```diff
@@ -190,6 +195,13 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
             # One smooth for all children when there isn't a child location.
             location = None
         model.random_effect[(smooth.rate, location)] = re_grid
+        random_effect_by_rate[smooth.rate].append(location)
+
+    for is_rate_consistent, locations in random_effect_by_rate.items():
```
alec-deason: `is_rate_consistent` is a very confusing name for this since it's just the name of the rate I think?

src/cascade/executor/construct_model.py
```diff
@@ -190,6 +195,13 @@ def construct_model_random_effects(default_age_time, single_age_time, ev_setting
...
+        if locations is not None and len(locations) != len(model.child_location):
```
alec-deason: Is this the test you want? Just the length and not set equality?

src/cascade/model/dismod_groups.py
```diff
@@ -70,19 +71,32 @@ def check_alignment(self, other):
...
-                    b_keys = [k[0] for k in b_keys]
-            a_keys = set(a_keys)
-            b_keys = set(b_keys)
-            if a_keys - b_keys:
-                one_not_other.append(f"left {group_name} has {a_keys - b_keys}")
-            elif b_keys - a_keys:
-                one_not_other.append(f"right {group_name} has {b_keys - a_keys}")
-            # else they agree
+            if group_name != "random_effect":
+                a_keys = list(group.keys())
```
alec-deason: Bit weird to coerce to a list and then immediately coerce to a set

src/cascade/executor/construct_model.py
adolgert: changed

src/cascade/model/dismod_groups.py
adolgert: removed the coercion and made it shorter.

PR #181, opened 2019-03-05
Feature/pass fits in memory

adolgert:
a) passes data down the drill
b) creates a main for dismodel_main.py
c) fixes bugs that cropped up when running manual tests with Liane's mvid
d) checks for "end" in the log

### DIFF
--- a/src/cascade/executor/estimate_location.py
+++ b/src/cascade/executor/estimate_location.py
@@ -23,6 +23,7 @@ from cascade.input_data.configuration.construct_country import (
 )
 from cascade.input_data.configuration.construct_mortality import get_raw_csmr, normalize_csmr
 from cascade.input_data.configuration.id_map import make_integrand_map
+from cascade.input_data.configuration.local_cache import LocalCache
 from cascade.input_data.configuration.raw_input import validate_input_data_types
 from cascade.input_data.db.asdr import asdr_as_fit_input
 from cascade.input_data.db.country_covariates import country_covariate_set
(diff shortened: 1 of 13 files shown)


src/cascade/input_data/configuration/local_cache.py
```diff
@@ -0,0 +1,26 @@
+from cascade.core import getLoggers
+
+CODELOG, MATHLOG = getLoggers(__name__)
+
+
+class LocalCache:
+    """A small in-memory cache object. This has the same interface
+    as memcached, so that we can switch it out for that, if appropriate."""
+    def __init__(self, maxsize=1):
```
alec-deason: maxsize=1 seems like a very odd default. Why? It doesn't turn the cache off but it also doesn't allow it to be useful.

src/cascade/executor/estimate_location.py
```diff
@@ -126,11 +130,11 @@ def retrieve_data(execution_context, local_settings, covariate_data_spec):
     data.study_id_to_name, data.country_id_to_name = find_covariate_names(
         execution_context, covariate_data_spec)
     # These are the draws as output of the parent location.
-    data.draws = None
+    data.draws = local_cache.get(f"fit-draws:{local_settings.grandparent_location_id}")
 
     # The parent can also supply integrands as a kind of prior.
     # These will be shaped like input measurement data.
-    data.integrands = None
+    data.integrands = local_cache.get(f"fit-integrands:{local_settings.grandparent_location_id}")
```
alec-deason: No matching set?

src/cascade/input_data/configuration/local_cache.py
adolgert: I increased it. It's meant for large things, and it's super dangerous b/c it will increase memory over time. This should be memcached, or go to a file, or explicitly purged. The 1 is a measure of my discomfort. :) It's now 8.

src/cascade/executor/estimate_location.py
adolgert: We aren't generating them yet, but that's a way to get them.

PR #189, opened 2019-03-07
Make samples run in parallel

alec-deason:


src/cascade/core/subprocess_utils.py
```diff
@@ -46,23 +47,27 @@ def inner(text):
...
-        await std_out_task
-        await std_err_task
+    return sub_process.returncode, "".join(stdout_accumulator), "".join(stderr_accumulator)
 
-        return sub_process.returncode, "".join(stdout_accumulator), "".join(stderr_accumulator)
 
-    return loop.run_until_complete(coroutine())
+def run_with_logging(command):
+    loop = asyncio.get_event_loop()
+    if loop.is_running():
```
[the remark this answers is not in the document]
alec-deason: Exactly

src/cascade/executor/estimate_location.py
```diff
@@ -277,18 +284,20 @@ def _fit_and_predict_fixed_effect_sample(sim_model, sim_data, fit_file, location
...
+                base_path / fit_file,
                 input_data.locations_df,
                 model.location_id,
                 local_settings,
                 draw_idx
             ))
 
         results = await asyncio.gather(*jobs)
-    return results
+    return [r for r in results if r is not None]
```
adolgert: I wouldn't have guessed that.

src/cascade/executor/estimate_location.py
```diff
@@ -277,18 +284,20 @@ def _fit_and_predict_fixed_effect_sample(sim_model, sim_data, fit_file, location
...
 
 
-def make_draws(model, input_data, max_fit, local_settings, num_processes):
+def make_draws(execution_context, model, input_data, max_fit, local_settings, num_processes):
+    base_path = execution_context.db_path(local_settings.parent_location_id)
+    base_path.mkdir(parents=True, exist_ok=True)
```
adolgert: Nice job pulling that in.

src/cascade/executor/estimate_location.py
alec-deason: Hmm? I have to return something from the jobs, even when the fail, so here I filter the failures out.

PR #190, opened 2019-03-07
added a list of what is not working

adolgert:
added list of what's not working. shocked to see I forgot to implement detailed ages and times.


docs/epiviz/overview.rst
```diff
@@ -3,13 +3,18 @@
...
 EpiViz-AT is a web page on http://internal.ihme.washington.edu/epi-at/.
 After you fill out
 that web page and hit submit, the web server runs a program on the cluster.
 We call that program ``dmcascade``. It is installed on the IHME cluster,
 and the code itself is stored at https://github.com/ihmeuw/cascade.
 This documentation, that you're reading, is at https://cascade.readthedocs.io.
 
+System Diagram
+^^^^^^^^^^^^^^
 The Cascade is composed of several systems, some of which are shown in
```
Brianepaulson: If its of use, here are summary definitions we worked through some time ago, should you want to use these:  
**EpiViz-AT** is the webpage where you define a model version, submit the job to the cluster and view results.  
**Cascade-AT** estimates incidence, prevalence and mortality rates for a single disease for every country or district in the world from the sum of all available disease and death observations in all countries.
**DisMod-AT** program enables a public health scientist to model the rate at which we get a condition, recover from a condition, die from a condition and die from other causes.  It is called Dismod-AT because the condition is often a disease and it is used to model the rates as stochastic functions of age and time.

docs/epiviz/overview.rst
```diff
@@ -30,3 +38,53 @@ This Cascade documentation looks at the problem in parts.
  *  Model Building - How the Cascade uses EpiViz-AT settings and the Cascade API.
  *  Command-line tools - Executable applications on the cluster, including :ref:`dmcascade-command-line`
  *  Operations and Maintenance - Processes to manage the Cascade.
+
+
+What Isn't Working Yet
```
Brianepaulson: A referral to a project HUB page that contains a Roadmap would be preferred over embedded comments that could go stale very quickly and confuse readers.  A HUB-roadmap, (where we could eventually manage feature target dates) would be the right place to set and manage expectations of functionality that is forthcoming.   LMK if you prefer a release note and roadmap section of this documentation, where I would update the content with each sprint end.

docs/epiviz/overview.rst
adolgert: added

docs/epiviz/overview.rst
adolgert: removed. You can add the link to the hub at any point. It would work better if it linked to images of the pages that showed what works and what doesn't, but that's a lot to ask, and I hope to get it to all-green soon.

PR #196, opened 2019-03-11
added logging to save predictions

adolgert:
This adds some logging. It doesn't solve the bug.

* adds logging of number of samples uploaded
* adds a num-samples argument on command line to make testing much faster.
* quiets a message on the MATHLOG about overwriting a file. That's not a problem really.

### DIFF
--- a/src/cascade/executor/cascade_plan.py
+++ b/src/cascade/executor/cascade_plan.py
@@ -97,6 +97,11 @@ class CascadePlan:
             sex_id = [self._settings.model.drill_sex, 3]
 
         policies = policies_from_settings(self._settings)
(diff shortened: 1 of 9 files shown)


src/cascade/executor/cascade_plan.py
```diff
@@ -97,6 +97,11 @@ def cascade_job(self, cascade_job_id):
             sex_id = [self._settings.model.drill_sex, 3]
 
         policies = policies_from_settings(self._settings)
+        if self._args.num_samples:
+            sample_cnt = self._args.num_samples
```
alec-deason: Yes, good

PR #198, opened 2019-03-11
This makes the main run on all fair nodes but unit tests have to be r…

adolgert:
…un on archive nodes.

I decided to make this happen because I realized that our limitation is for testing, not running. The patch below means we can run on any node, but I didn't fix the fussy parts that would allow us to test on any node.

### DIFF
--- a/docs/epiviz/file_locations.rst
+++ b/docs/epiviz/file_locations.rst
@@ -1,0 +1,27 @@
+.. _file-locations:
+
+File Locations on the Cluster
+=============================
+
+ *  ``/ihme/epi/dismod_at``
+
+    *  ``share/local_odbc.ini`` - This is an ODBC file to use when running on
+       the Fair cluster. This is used in ``dismodel_main.py`` while
+       their password technique involves using the unavailable J drive.
+
+    *  ``bin/`` - Contains tools for modelers to manipulate bundles on
+       the debug database. Subdirectories of this contain hacked copies
+       of the uploader.
+
+    *  ``log/`` - The MATHLOG, by model version id, is in this directory.
+
+    *  ``runscripts/`` - The shell scripts run by EpiViz-AT are here.
+       They are copied from the scripts directory in the Github.
+
+ *  ``/ihme/epi/at_cascade``
+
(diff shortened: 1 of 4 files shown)


docs/epiviz/file_locations.rst
```diff
@@ -0,0 +1,27 @@
...
+File Locations on the Cluster
+=============================
+
+ *  ``/ihme/epi/dismod_at``
+
+    *  ``share/local_odbc.ini`` - This is an ODBC file to use when running on
+       the Fair cluster. This is used in ``dismodel_main.py`` while
+       their password technique involves using the unavailable J drive.
+
+    *  ``bin/`` - Contains tools for modelers to manipulate bundles on
```
alec-deason: If we're formalizing this :hankey: we should move the elmo environment out of my personal directory.

src/cascade/core/db.py
```diff
@@ -63,6 +65,18 @@ def __dir__(self):
...
+def use_local_odbc_ini():
+    """The password vault is an odbc.ini file that's on a drive that
+    isn't accessible from all nodes, so copy it to a place it
+    can be found and point the connection generation to it."""
+    have_default = Path("/home/j").exists() or Path("~/.odbc.ini").expanduser().exists()
+    if LOCAL_ODBC.exists() and not have_default:
+        CODELOG.info(f"Using odbc.ini at {LOCAL_ODBC}")
+        db_tools.config.DBConfig(odbc_filepath=str(LOCAL_ODBC))
+    else:
+        CODELOG.debug(f"Using default odbc.ini")
```
alec-deason: Should there be a third branch here where we blow up if there's no default and also no J? It'll happen on it's own later but... hm, no, you really do want to lazily fail because you might be doing something that doesn't require the databases. Ok. Never mind.

docs/epiviz/file_locations.rst
adolgert: That might be work. The hacked Elmo is what we have now.

alec-deason: Oh, I'm not suggesting we move away from the hacked elmo just that it shouldn't live in a place with my username on it.

PR #200, opened 2019-03-12
Enhancement/improve testing around uploader

alec-deason:


src/cascade/saver/save_prediction.py
```diff
@@ -1,29 +1,47 @@
...
+    predicted["sex_id"] = predicted.s_sex.apply(lambda s_sex: {-1: 2, 0: 3, 1: 1}[int(s_sex * 2)])
     predicted = predicted.drop("s_sex", "columns")
 
     predicted = age_ranges_to_groups(execution_context, predicted)
 
     integrand_to_measure = {v.name: k for k, v in make_integrand_map().items()}
+    # TODO: This should actually be fixed in id_map but I'm not sure how best to do that so I'm punting
+    integrand_to_measure[IntegrandEnum.prevalence.name] = 5
+    integrand_to_measure[IntegrandEnum.Tincidence.name] = 42
+    integrand_to_measure[IntegrandEnum.Sincidence.name] = 41
```
adolgert: Ohhhh. Yes, this belongs elsewhere. The problem is that the inverse isn't unique. I can add a column to flag which of them should be the unique inverse and a test to ensure uniqueness. Not this PR, but good!

src/cascade/saver/save_prediction.py
```diff
@@ -1,29 +1,47 @@
...
-    elif fit_or_final == "final":
-        table = "model_estimate_final"
-    else:
-        raise ValueError("fit_or_final must be 'fit' or 'final'")
+def uncertainty_from_prediction_draws(predictions):
+    predictions = pd.concat(predictions)
+    columns_to_remove = ["sample_index"] + [c for c in predictions.columns if c.startswith("s_") and c != "s_sex"]
+    predictions = predictions.drop(columns_to_remove, "columns")
+    predictions = predictions.groupby(
+        ["location", "integrand", "age_lower", "age_upper", "time_lower", "time_upper", "s_sex"]
```
adolgert: I'm not seeing changes here. Just moved to a better spot.

tests/input_data/db/test_demographics.py
```diff
@@ -0,0 +1,36 @@
...
+from cascade.input_data.db.demographics import age_ranges_to_groups
+
+
+def test_age_ranges_to_groups(mocker):
+    fake_get_age_groups = mocker.patch("cascade.input_data.db.demographics.get_age_groups")
+    fake_get_age_groups.return_value = pd.DataFrame(
+        {"age_group_id": [1, 2, 3, 4], "age_group_years_start": [0, 1, 25, 90], "age_group_years_end": [1, 2, 30, 100]}
+    )
+
+    df = pd.DataFrame({"age_lower": [0, 1, 90], "age_upper": [1, 2, 100], "integrand": ["remission"] * 3})
```
adolgert: Good. This tests the inverse problem. That part always bothered me. It tests the age part.

tests/saver/test_save_prediction.py
```diff
@@ -0,0 +1,105 @@
...
+    locations = [1, 2]
+    age_lower, age_upper = zip(*[(0, 0.5), (50, 60)])
+    sexes = [-0.5]
+    time_lower, time_upper = zip(*[(y, y) for y in [2000, 2020]])
+    integrands = ["prevalence", "Sincidence"]
+
+    df = pd.DataFrame(
+        index=pd.MultiIndex.from_product(
+            [locations, age_lower, age_upper, time_lower, time_upper, integrands, sexes],
+            names=["location", "age_lower", "age_upper", "time_lower", "time_upper", "integrand", "s_sex"],
```
adolgert: Man, these multiindex are so complicated.

tests/saver/test_save_prediction.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+    ).reset_index()
+
+    true_mean = np.linspace(0.0001, 2.0, len(df))
+    rng.shuffle(true_mean)
+    true_std = np.linspace(0.00001, 0.001, len(df))
+    rng.shuffle(true_std)
```
adolgert: You're using rng!

tests/saver/test_save_prediction.py
```diff
@@ -0,0 +1,105 @@
...
+    ).sort_index()
+    source["lower"] = source[["true_mean", "true_std"]].apply(
+        lambda r: norm(r["true_mean"], r["true_std"]).ppf(0.025), axis="columns"
+    )
+    source["upper"] = source[["true_mean", "true_std"]].apply(
+        lambda r: norm(r["true_mean"], r["true_std"]).ppf(0.975), axis="columns"
+    )
+
+    assert np.allclose(with_uncertainty["lower"], source["lower"], rtol=0.04)
+    assert np.allclose(with_uncertainty["upper"], source["upper"], rtol=0.1)
```
adolgert: These are fine. Yes, they seem wide.

tests/saver/test_save_prediction.py
```diff
@@ -0,0 +1,105 @@
...
+    fake_age_to_group = mocker.patch("cascade.saver.save_prediction.age_ranges_to_groups")
+    fake_age_to_group.side_effect = lambda _, df: df
+
+    predicted = pd.DataFrame(
+        {
+            "mean": [1, 1, 1],
+            "location": [1, 2, 3],
+            "integrand": ["prevalence", "Sincidence", "remission"],
+            "age_lower": [0, 1, 2],
+            "age_upper": [1, 2, 3],
```
adolgert: Would it be hard to copy this test to make one that excludes, say, `sex_id=1`? And ensure it doesn't end up in the result?

src/cascade/executor/estimate_location.py
```diff
@@ -256,21 +256,7 @@ def compute_location(execution_context, local_settings, input_data, model):
...
-
-    lower = predictions.quantile(0.025)
-    lower.columns = ["lower"]
-    upper = predictions.quantile(0.975)
-    upper.columns = ["upper"]
-    mean = predictions.mean()
-    mean.columns = ["mean"]
-
-    predictions = pd.concat([lower, upper, mean], axis="columns").reset_index()
+    predictions = uncertainty_from_prediction_draws(predictions)
```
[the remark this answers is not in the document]
alec-deason: I feel like it's clearer to keep the name when I have a series of transformations on what in conceptually the same data. And when I'm not going to use the intermediates for anything. It's an alternative to just composing all the transformations together.

src/cascade/saver/save_prediction.py
alec-deason: The same pattern appears in the sqlite->uploader spreadsheet code for the same reason

tests/saver/test_save_prediction.py
alec-deason: If you have a cleaner way of constructing this dataframe I'm all ears. I could do it as a big loop but I think that's actually more distracting from the test and this trick (and it's totally a dumb trick) is the canonical way of doing this as far as I know.

tests/saver/test_save_prediction.py
alec-deason: I want a local seed! Also this is the pattern rust encourages so I've gotten used to it.

tests/saver/test_save_prediction.py
alec-deason: They're wide but the numbers involved are very small. As I increased the number of samples I actually had to loosen these to accommodate the points with the lowest std. I don't really have a sense for how reasonable that is.

tests/saver/test_save_prediction.py
alec-deason: Yep, this is clearly a case where some parametrization would be good. I'll do that.

src/cascade/executor/estimate_location.py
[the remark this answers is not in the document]
adolgert: OK. Got it. This was a kind-of from me, so that works.

tests/saver/test_save_prediction.py
adolgert: My method for Pandas is to review the docs on the data structure, make a stab, and test it. Anything that passes tests is going to have to be good enough.

tests/saver/test_save_prediction.py
adolgert: I'm happy about the rng use. That's what I meant to say. And seeding is the right thing.

tests/saver/test_save_prediction.py
adolgert: Hmmm. The math looks right to me. These stats tests are difficult.

PR #201, opened 2019-03-12
Feature/detailed priors on grids

adolgert:
This enables a modeler to

* select mulstd priors on the Rate, Random Effect, Study Covariate, and Country Covariate tabs
* create detailed priors on the Rate, Random Effect, Study Covariate, and Country Covariate tabs

The second feature is more likely to be useful.

### DIFF
--- a/src/cascade/executor/construct_model.py
+++ b/src/cascade/executor/construct_model.py
@@ -190,14 +190,16 @@ def matching_knots(rate_grid, smoothing_prior):
     extents = dict()
     for extent in ["age", "time", "born"]:
         extents[extent] = np.zeros(2, dtype=np.float)
-        for sidx, side, default_extent in [(0, "lower", -inf), (1, "upper", inf)]:
+        for side_idx, side, default_extent in [(0, "lower", -inf), (1, "upper", inf)]:
             name = f"{extent}_{side}"
             if smoothing_prior.is_field_unset(name):
-                extents[extent][sidx] = default_extent
+                extents[extent][side_idx] = default_extent
             else:
-                extents[extent][sidx] = getattr(smoothing_prior, name)
+                extents[extent][side_idx] = getattr(smoothing_prior, name)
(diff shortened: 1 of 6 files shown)


src/cascade/executor/construct_model.py
```diff
@@ -147,16 +148,65 @@ def constraint_from_rectangular_data(rate_var, default_age_time):
...
+        smoothing_prior (cascade.input_data.configuration.form.SmoothingPrior):
+            A single smoothing prior.
+
+    Returns:
+        Iterator over (a, t) that match. Can be nothing.
+    """
+    extents = dict()
+    for extent in ["age", "time", "born"]:
+        extents[extent] = np.zeros(2, dtype=np.float)
+        for sidx, side, default_extent in [(0, "lower", -inf), (1, "upper", inf)]:
```
alec-deason: I stumbled over `sidx`. I think `s_idx` or `side_idx` would have been clear to me.

src/cascade/executor/construct_model.py
```diff
@@ -147,16 +148,65 @@ def constraint_from_rectangular_data(rate_var, default_age_time):
...
+            name = f"{extent}_{side}"
+            if smoothing_prior.is_field_unset(name):
+                extents[extent][sidx] = default_extent
+            else:
+                extents[extent][sidx] = getattr(smoothing_prior, name)
+    ages, times = np.meshgrid(rate_grid.ages, rate_grid.times)
+    # result is shape (len(times), len(ages)), backwards
```
alec-deason: I have no idea what this means

src/cascade/executor/construct_model.py
adolgert: That helps.

src/cascade/executor/construct_model.py
adolgert: That helps, too. I can express this more clearly. This will be read more times than it was written.

PR #213, opened 2019-03-19
Feature/fixed before random policy

alec-deason:
This splits "estimate_location" into several sequential steps and then adds one optional step in the middle of the sequence to calculate an initial guess  from a run of fit fixed

### DIFF
--- a/src/cascade/executor/cascade_plan.py
+++ b/src/cascade/executor/cascade_plan.py
@@ -227,16 +227,23 @@ class CascadePlan:
             setup_task = []
         else:
             setup_task = [(drill[0], "bundle_setup")]
+
+        substeps = [
+            "prepare_data",
+            "construct_model",
+        ]
+        if settings.policies.fit_strategy == "fit_fixed_then_fit":
+            substeps.append("initial_guess_from_fit_fixed")
+        substeps.extend([
+            "compute_initial_fit",
+            "compute_draws_from_parent_fit",
+            "save_predictions"
+        ])
+
         tasks = setup_task + [
             (drill_location, ("estimate_location", substep))
             for drill_location in drill
-            for substep in [
-                "prepare_data",
-                "construct_model",
-                "compute_initial_fit",
-                "compute_draws_from_parent_fit",
-                "save_predictions"
-            ]
+            for substep in substeps
         ]
         task_pairs = list(zip(tasks[:-1], tasks[1:]))
         plan._task_graph = nx.DiGraph()
--- a/src/cascade/executor/cascade_plan.py
+++ b/src/cascade/executor/cascade_plan.py
@@ -136,10 +136,10 @@ class CascadePlan:
         so a drill starting halfway will not have a grandparent location.
         There are child locations for the last task though.
         """
-        parent_task = list(self._task_graph.in_edges(cascade_job_id))
-        if parent_task and self._job_kind(parent_task[0][0]) == "estimate_location":
-            # [only edge][(edge start, edge finish)][(location, index)]
-            grandparent_location_id = self._location_of_cascade_job(parent_task[0][0])
+        parent_location_id = self._location_of_cascade_job(cascade_job_id)
+        predecessors = list(self._locations.predecessors(parent_location_id))
+        if predecessors:
+            grandparent_location_id = predecessors[0]
         else:
             grandparent_location_id = None
 
@@ -227,7 +227,11 @@ class CascadePlan:
             setup_task = []
         else:
             setup_task = [(drill[0], "bundle_setup")]
-        tasks = setup_task + [(drill_location, 0) for drill_location in drill]
+        tasks = setup_task + [
+            (drill_location, ("estimate_location", substep))
+            for drill_location in drill
+            for substep in ["prepare_data", "construct_model", "compute_and_save_draws"]
+        ]
         task_pairs = list(zip(tasks[:-1], tasks[1:]))
         plan._task_graph = nx.DiGraph()
         plan._task_graph.add_nodes_from(tasks)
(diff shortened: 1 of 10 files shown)


src/cascade/executor/cascade_plan.py
```diff
@@ -227,7 +227,24 @@ def from_epiviz_configuration(cls, locations, settings, args):
...
+            "compute_initial_fit",
+            "compute_draws_from_parent_fit",
+            "save_predictions"
+        ])
+
+        tasks = setup_task + [
+            (drill_location, ("estimate_location", substep))
+            for drill_location in drill
+            for substep in substeps
+        ]
```
adolgert: The line after this, the `task_pairs`, is where this falls down. We need to preserve, in the graph, the dependency tree, so that we can know what can run in parallel.

src/cascade/executor/dismodel_main.py
```diff
@@ -51,16 +55,26 @@ def main(args):
...
     for cascade_task_identifier in plan.cascade_jobs:
         cascade_job, this_location_work = plan.cascade_job(cascade_task_identifier)
         configure_execution_context(execution_context, args, this_location_work)
 
         if cascade_job == "bundle_setup":
             # Move bundle to next tier
             setup_tier_data(execution_context, this_location_work.data_access, this_location_work.parent_location_id)
-        elif cascade_job == "estimate_location":
-            estimate_location(execution_context, this_location_work, local_cache)
+        elif cascade_job == "estimate_location:prepare_data":
```
adolgert: This is a big list. Totally makes sense. Looks a little like an `eval()`, doesn't it?

src/cascade/executor/cascade_plan.py
adolgert: I see now that these are, in fact, linear, so the zip is fine for this specific application. Interesting. The ` `compute_draws` should be the task array, but... does it show up as one job here? I'm currently still confused about that.

src/cascade/executor/estimate_location.py
```diff
@@ -52,24 +51,68 @@ def estimate_location(execution_context, local_settings, local_cache=None):
...
+    local_cache.set("covariate_data_spec:{local_settings.parent_location_id}", covariate_data_spec)
     input_data = retrieve_data(execution_context, local_settings, covariate_data_spec, local_cache)
     columns_wrong = validate_input_data_types(input_data)
     assert not columns_wrong, f"validation failed {columns_wrong}"
     modified_data = modify_input_data(input_data, local_settings, covariate_data_spec)
+    local_cache.set("prepared_input_data:{local_settings.parent_location_id}", modified_data)
+
+
+def construct_model_for_estimate_location(local_settings, local_cache):
+    covariate_multipliers = local_cache.get("covariate_multipliers:{local_settings.parent_location_id}")
```
adolgert: I don't like this get/set on the cache at all. Looks like we don't know how to write a function that has inputs and outputs. Maybe these functions should look like non side-effecting functions, and it's outside of these that we serialize and deserialize what they need.

adolgert: Oh, duh. This is the outside part. I see. And you've written the serialization by hand, which is an excellent first pass.

src/cascade/executor/cascade_plan.py
alec-deason: Yes, they are linear. I haven't touched the problem of how to represent dependency. Nor have I broken compute_draws into individual jobs. It's currently a single job like it's always been. The reason for that is that breaking it up such that we preserve parallelism within a single-node job requires a lot of extra work and I decided that was out of scope.

src/cascade/executor/dismodel_main.py
alec-deason: Yeah, this is an ugly sieve. I'm not sure exactly how to structure it more cleanly but you could clearly do better.

alec-deason: You could make the arguments list homogeneous and then just do a dictionary look up to get the handler for a particular job name. Or have the reference to the handler as a parameter on the job, though that changes other things about the architecture.

src/cascade/executor/estimate_location.py
alec-deason: Yeah, I agree this is ugly. You need _some_ way to annotate which values are passed through the cache. Could be function argument names and a returned dictionary where the outer wrapper takes care of packing and unpacking, that's more fiddly but makes this code look cleaner. Or you can do something like this where you explicitly pack and unpack values. I don't know what the best answer is. The explicit packing and unpacking doesn't bother me particularly. What does bother me is that, as I wrote this out, it felt like I was sticking stuff into the cache pretty arbitrarily, which I was. I think you could sit and think a little more carefully about how this get's partitioned and sequenced and make it feel more orderly.

adolgert: These messages are part of a protocol. In order to know bandwidth, to ensure there isn't crazy cache use, I like to have a list of what those things are or will be. Dynamic is good, but this is a nice place to put a layer. I'll think about it. Meanwhile, what you have is a good step.

tests/executor/test_cascade_plan.py
```diff
@@ -116,11 +117,11 @@ def test_model_options(loc, expected):
     bound_form = MiniForm(dict(
         model=dict(bound_random=0.7),  # This will be the default value.
     ))
-    bound_form.re_bound_location=[
```
adolgert: How did this no-space equals even get in there?

tests/executor/test_construct_model.py
```diff
@@ -98,7 +98,7 @@ def make_local_settings(given_settings):
     j = list(c.cascade_jobs)[1:]
     job_choice = choices.choice(list(range(len(j))), name="job_idx")
     job_kind, job_args = c.cascade_job(j[job_choice])
-    assert job_kind == "estimate_location"
+    assert job_kind.startswith("estimate_location:")
```
adolgert: This feels like an encoding that should be abstracted.

alec-deason: Or done away with. It's there mostly because I wanted to keep my changes fairly minimal. I'm not sure there will even be different classes of jobs other than "estimate_location".

PR #214, opened 2019-03-21
code for creating a new model version id from an existing settings file

alec-deason:
Code to create a new model version id from existing settings, including uploading a copy of the settings.

### DIFF
--- a/src/cascade/saver/create_model_version.py
+++ b/src/cascade/saver/create_model_version.py
@@ -1,0 +1,77 @@
+import json
+
+from cascade.core import getLoggers, __version__
+from cascade.core.db import cursor
+
+CODELOG, MATHLOG = getLoggers(__name__)
+
+
+def create_model_version(execution_context, settings, username, title=None):
+    new_mvid = _insert_model_version(execution_context, settings, username, title)
+
+    _insert_settings_json(execution_context, settings, new_mvid)
+
+    MATHLOG.info(f"Created new model version: {new_mvid}")
+
+
+def _insert_model_version(execution_context, settings, username, title):
+    if title is None:
+        title = settings.model.title
+
+    query = """
+    insert into model_version
+    (
+        modelable_entity_id,
+        bundle_id,
+        single_measure_id,
+        title,
+        gbd_round_id,
+        csmr_cod_output_version_id,
+        csmr_mortality_output_version_id,
+        inserted_by,
+        model_version_status_id,
+        code_version,
+        location_set_version_id
+    )
+    values (
+        %(modelable_entity_id)s,
+        %(bundle_id)s,
+        null,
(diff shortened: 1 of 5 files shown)


src/cascade/saver/create_model_version.py
```diff
@@ -0,0 +1,77 @@
...
+                gbd_round_id=settings.gbd_round_id,
+                cod_version_id=settings.csmr_cod_output_version_id,
+                # TODO: We don't distinguish this internally though theoretically they could differ
+                mortality_version_id=settings.csmr_cod_output_version_id,
+                code_version=__version__,
+                location_set_version_id=settings.location_set_version_id,
+                username=username,
+            ),
+        )
+        return c.lastrowid
```
adolgert: That's it, huh? It's the last row ID?

src/cascade/saver/create_model_version.py
```diff
@@ -0,0 +1,77 @@
+import json
+
+from cascade.core import getLoggers, __version__
+from cascade.core.db import cursor
+
+CODELOG, MATHLOG = getLoggers(__name__)
```
adolgert: Do we make it in the saver or at the start of the model? It might be at the start because the rest of the machinery assumes an mvid and uses that to decide where to save files in a separate directory.

src/cascade/saver/create_model_version.py
alec-deason: That's how autoincrement keys do.

src/cascade/saver/create_model_version.py
alec-deason: I was imagining it at the end but you're totally right, the tier 3 promotion code needs an mvid and that's at the beginning.

alec-deason:
There are actually several things this does not do which the original from Epiviz does. It doesn't pull the current mortality or location versions or the current gbd round id. It gets all of those out of the incoming settings. I think that's the right thing to do, especially since all those versions are (I think) slow changing. But you could add the extra queries in pretty easily. The current goo for all that is here:  https://stash.ihme.washington.edu/projects/VIZ/repos/epi/browse/php/api/model/save.php?at=refs%2Fheads%2FAT%2Fdev

src/cascade/saver/create_model_version.py
adolgert: fair enough

src/cascade/saver/create_model_version.py
adolgert: That's fine. We can put this code wherever when it's clearer how to make the `main()` that they can use. It's great to have it.

alec-deason: Yeah, can't have this in the main version. It's something they have to do deliberately.

PR #220, opened 2019-03-27
moving to 37 if tools allow

adolgert:
We have been using python 3.7 but haven't changed requirements to reflect this. We have a real dependency on python 3.7 because we use the "async" keyword when starting subprocesses.


mbannick:
Note: `elmo` is not working with Python 3.7, so will stick to Python 3.6 for now.

PR #257, opened 2019-10-10
DISMODAT-544: Pecking order for bundle data / measurement standard error

mbannick:
Need to implement a pecking order for what should be used for standard error if standard error is *not* present in the bundle downloaded from the epi database.

Currently, uncertainty intervals over-ride the standard error. We are changing this so that the order is:
1. Standard error
2. Uncertainty intervals
3. Effective sample size
4. Sample size

This change does not apply to data from, for example, the age-standardized death rate inputs or cause-specific mortality inputs (still use just `bounds_to_stdev` or `meas_bounds_to_stdev` in those cases).


src/cascade/stats/estimation.py
```diff
@@ -31,6 +31,47 @@ def meas_bounds_to_stdev(df):
...
+        mean_value:
+        es:
+
+    Returns:
+
+    """
+    mean = mean_value.copy()
+    sample_size = es.copy()
+    count = mean * sample_size
+
```
gma-uw: Don't need to copy the input arguments if they are unmodified by the function.

src/cascade/stats/estimation.py
```diff
@@ -31,6 +31,47 @@ def meas_bounds_to_stdev(df):
...
+    std_5 = np.sqrt(5.0 / sample_size**2)
+
+    std = np.sqrt(mean / sample_size)
+
+    # For counts < 5, linearly interpolate between std_0 and std_5,
+    # replacing the regular standard deviation.
+    std[under_5] = ((5.0 - count[under_5]) * std_0[under_5] +
+                    count[under_5] * std_5[under_5]) / 5.0
+    return std
+
```
gma-uw: I think the calculations are correct and yours is more clear than mine.  Perhaps we should just implement the Wilson code, using a keyword argument,  proportional = False, then we will not have to revisit this in the future, just change the calls to the function.

Also we should find out if we have the proportional info available from anywhere.

src/cascade/stats/estimation.py
```diff
@@ -31,6 +31,47 @@ def meas_bounds_to_stdev(df):
     return df.drop(["meas_lower", "meas_upper"], axis=1)
 
 
+def es_to_stdev(mean_value, es):
+    """
+    Takes an array of values (like mean), and
+    effective sample size (es) and transforms it
+    to standard error assuming that the sample
+    size is Poisson distributed.
```
gma-uw: Actually the small sample interpolates between binomial and Poisson.

src/cascade/stats/estimation.py
```diff
@@ -47,3 +88,61 @@ def bounds_to_stdev(lower, upper):
         pd.Series: The standard deviation for that CI.
     """
     return (upper - lower) / (2 * 1.96)
+
+
```
gma-uw: Your approach of computing all the options and masking them into the result is more clear than my masked calculations were. I like yours.

src/cascade/stats/estimation.py
```diff
@@ -47,3 +88,61 @@ def bounds_to_stdev(lower, upper):
...
+    # Calculate standard deviation different ways (also
+    stdev_from_bounds = bounds_to_stdev(lower=df.lower.values, upper=df.upper.values)
+    stdev_from_es = es_to_stdev(mean_value=df.mean.values, es=df.effective_sample_size.values)
+
+    # Use boolean arrays representing the pecking order to replacing standard error
+    standard_error[replace_se_with_ui] = stdev_from_bounds[replace_se_with_ui]
+    standard_error[replace_se_with_es] = stdev_from_es[replace_se_with_es]
+
+    # Do a final check on standard error
```
gma-uw: I am wondering if we should allow std's == 0 to pass through, and let dismod fix them with minCV by integrand. If we decide to do that, we should catch the case where both mean == std == 0. That will stop dismod.

src/cascade/stats/estimation.py
mbannick: - [ ] Wilson's Score Interval

src/cascade/stats/estimation.py
mbannick: - [ ] Don't copy values

src/cascade/stats/estimation.py
mbannick: I think this is something to flag with the team. Left a comment on the JIRA ticket DISMODAT-544 for us to figure out.

src/cascade/stats/estimation.py
```diff
@@ -31,44 +32,60 @@ def meas_bounds_to_stdev(df):
...
 
 
-def es_to_stdev(mean_value, es):
+def wilson_interval(prop, ess):
+    """
+    Calculate Wilson Score Interval
+    Args:
+        prop: proportion
+        ess: effective sample size
+
```
gma-uw: Note: q=0.975  is for a standard normal 95% confidence interval.

PR #259, opened 2019-10-17
Added a temporary patch to compute reference covariates from child av…

gma-uw:
…erages if the parent reference is not available. If neither are available, throw an exception.

### DIFF
--- a/src/cascade/executor/covariate_data.py
+++ b/src/cascade/executor/covariate_data.py
@@ -69,7 +69,29 @@ def add_country_covariate_to_observations_and_avgints(data, local_settings, epiv
     country_specs = {ccov for ccov in epiviz_covariates if ccov.study_country == "country"}
     for covariate_id in {evc.covariate_id for evc in country_specs}:
         ccov_ranges_df = data.country_covariates[covariate_id]
-        reference = reference_value_for_covariate_mean_all_values(ccov_ranges_df)
+
+        parent_ccov_ranges = ccov_ranges_df[ccov_ranges_df.location_id.isin([local_settings.parent_location_id])]
+        if not parent_ccov_ranges.empty:
+            CODELOG.warning ("****** FIXME -- The parent had ccov values -- population weight these *********")
+            reference = reference_value_for_covariate_mean_all_values(parent_ccov_ranges)
+        else:
+            __patch__ = True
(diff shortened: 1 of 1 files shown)


src/cascade/executor/covariate_data.py
```diff
@@ -69,7 +69,29 @@ def add_country_covariate_to_observations_and_avgints(data, local_settings, epiv
     country_specs = {ccov for ccov in epiviz_covariates if ccov.study_country == "country"}
     for covariate_id in {evc.covariate_id for evc in country_specs}:
         ccov_ranges_df = data.country_covariates[covariate_id]
-        reference = reference_value_for_covariate_mean_all_values(ccov_ranges_df)
+
+        parent_ccov_ranges = ccov_ranges_df[ccov_ranges_df.location_id.isin([local_settings.parent_location_id])]
+        if not parent_ccov_ranges.empty:
+            CODELOG.warning ("****** FIXME -- The parent had ccov values -- population weight these *********")
```
mbannick: Can you remove this whitespace? Travis automatically checks for pep8 formatting, and it will say our build is failing if we merged this.

src/cascade/executor/covariate_data.py
```diff
@@ -69,7 +69,29 @@ def add_country_covariate_to_observations_and_avgints(data, local_settings, epiv
...
+            if __patch__:
+                # FIXME -- Below is a temporary patch -- all locations should have covariate references
+                CODELOG.error(f"Country covariates for location {local_settings.parent_location_id} were missing.")
+                child_ccov_ranges = ccov_ranges_df[ccov_ranges_df.location_id.isin(local_settings.children)]
+                if not child_ccov_ranges.empty:
+                    missing = set(local_settings.children) - set(child_ccov_ranges.location_id)
+                    if missing:
+                        CODELOG.error(f"Country covariates for child locations {sorted(missing)} were missing.")
+                    else:
+                        CODELOG.warning ("****** FIXME -- The children had ccov values -- population weight these *********")
```
mbannick: Same here

src/cascade/executor/covariate_data.py
gma-uw: You have to be kidding. But I'll take them out.

PR #260, opened 2019-10-18
The call to griddata used pandas dataframe arguments, which apparentl…

gma-uw:
…y don't work in griddata. Passing numpy arrays instead.


gma-uw:
I am aborting this PR.

PR #263, opened 2019-10-18
Drew's dismod database interface code was retaining old values from r…

gma-uw:
…un to run,

which left table values from previous runs initialized to perhaps erroneous data.
Deleting the dismod database clears that information and starts fresh.

This was confusing, at best, and could possibly lead to incompatible data in the tables.

### DIFF
--- a/src/cascade/model/object_wrapper.py
+++ b/src/cascade/model/object_wrapper.py
@@ -52,6 +52,7 @@ class ObjectWrapper:
         Args:
             filename (Path|str|None): Path to filename or None if this
                 DismodFile should be in memory, as used for testing.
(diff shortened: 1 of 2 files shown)


src/cascade/model/object_wrapper.py
```diff
@@ -47,7 +47,7 @@ class ObjectWrapper:
     within those tables into higher-level objects that are then
     easier to reason about.
     """
-    def __init__(self, filename):
+    def __init__(self, filename, unlink=False):
```
mbannick: @gma-uw Can you add some info in the docstring about what `unlink` argument will do?

PR #267, opened 2019-10-21
The modeler should specify either Sincidence and Tincidence , but if …

gma-uw:
…measured data is marked as incidence, change it to Sincidence as that is easier for dismod to fit.


mbannick:
Should we definitely do this @gma-uw ? Or should we have the code fail if someone passes `Incidence`?

gma-uw:
I think we should do this -- perhaps we should have EPIVIS catch the problem earlier though.

On Oct 21, 2019, at 9:46 AM, Marlena Bannick <notifications@github.com<mailto:notifications@github.com>> wrote:


Should we definitely do this @gma-uw<https://github.com/gma-uw> ? Or should we have the code fail if someone passes Incidence?

PR #292, opened 2020-06-25
Set age/time bounds to be midpoint

jizezhang:
* set age time bounds to be midpoints
* make demographic conversion optional


mbannick:
@jizezhang this looks great! could you also please add a test for when `midpoint = True`?

PR #299, opened 2020-07-11
DISMODAT-693: Saving uncertainty in the prior

mbannick:
This PR will make it so that the prior has mean upper and lower saved, rather than just mean as it is now. In order to do this, I needed to modify the `rvs` function (random variates) for each of the prior distributions (in `priors.py`) so that it censored the distribution rather than just resampling until something fell within bounds. Then I get the upper and lower based on the prior `SmoothGrid`, and save that to the epi database using `ResultsHandler()` in the `dismod_db` executable.

Importantly, this will work the same way for priors starting at the top level of the cascade (global) as it will for going down the cascade where the priors are filled in by the fit from its parent.

### DIFF
--- a/src/cascade_at/cascade/cascade_stacks.py
+++ b/src/cascade_at/cascade/cascade_stacks.py
@@ -175,7 +175,8 @@ def root_fit(model_version_id: int, location_id: int, sex_id: int,
         both=False,
         predict=True,
         upstream_commands=upstream,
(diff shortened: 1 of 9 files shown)


mbannick:
@benmiltz not sure why it won't let me tag you as a reviewer, but was thinking reviewing this PR will help you get familiar with some of the recent changes

PR #301, opened 2020-08-20
DISMODAT-698: Upgrade to Jobmon Guppy

mbannick:
This PR upgrades to using the Guppy release for Jobmon.

1. Each `CascadeOperation` can make a `TaskTemplate` if it's passed a `Tool`, and then has a `command_template` that looks like this
```
{script} {model_version_id} ...
```
2. Each `CascadeOperation` has a `template_kwargs` attribute which looks like the following, reflecting the arguments for this instance of the cascade operation, and can then be passed to `TaskTemplate.create_task()` as keyword arguments.
```
{model_version_id: --model-version-id 101010}
```
3. The way that we convert a `CascadeOperation` into a `Task` is through passing the `template_kwargs` in (2) plus the script name into the `TaskTemplate` that looks like (1).
4. Incorporated workflow updates in `cascade_at.executor.run` and `cascade_at.
5. Kept the original `CascadeOperation.command` attribute so that we can still use this without jobmon locally (but hopefully can get rid of this when jobmon is open sourced)

Note: The scripts are installed as entry points so they are not absolute paths to scripts, and we don't call `python ...`. Instead the jobmon task commands should look like e.g. `run --model-version-id 0`.


src/cascade_at/executor/args/args.py
```diff
@@ -10,93 +10,109 @@ class StaticArgError(CascadeArgError):
...
     def __init__(self, arg=None):
 
-        self._arg = None
-        self._kwargs = dict()
+        self._flag = None
+        self._parser_kwargs = dict()
         if arg is not None:
-            self._arg = arg
+            self._flag = arg
+        self._task_arg = False
```
davidshaw2018, line 28: So all args are node_args by default? Is this intentional or am I missing somewhere task_args are set later?

src/cascade_at/jobmon/workflow.py
```diff
@@ -1,51 +1,51 @@
...
-    """
-    def __init__(self, upstream_commands=None, **kwargs):
-        super().__init__(**kwargs)
-        if upstream_commands is None:
-            upstream_commands = []
-        self.upstream_commands = upstream_commands
-        LOG.info(f"Created task with command {self.command} with "
-                 f"{len(self.upstream_commands)} upstream commands.")
+class JobmonConstants:
+    EXECUTOR = "SGEExecutor"
```
megan-h-moore, line 18: You will also want to import whichever executor(s) you are going to use otherwise I think you will run into an error. For SGEExecutor `from jobmon.client.execution.strategies.sge import SGEExecutor`

megan-h-moore (approved):
Looks fantastic! Glad those Tool commands seemed to plug-in smoothly.

src/cascade_at/executor/args/args.py
mbannick: Yep! They're all `node_args`. The only thing that defines data flow is `model_version_id` so that's the only `task_arg`

PR #305, opened 2020-09-01
Dismod spec says the avgint weight column should be nullable.

gma-uw:
Dismod spec says the avgint weight_id column should be nullable.


mbannick:
Thanks, @gma-uw !

PR #310, opened 2020-09-11
Test dismod

gma-uw:
There are 2 failing tests in this suite of tests, and they should fail because dismod doesn't seem to be working right.


tests/dismod_execution/dismod_tests.py
```diff
@@ -0,0 +1,196 @@
...
+    priors = dict(parent_density = 'uniform',
+                  parent_std = None,
+                  child_density = 'uniform',
+                  child_std = None,
+                  subgroup_density = iota_subgroup_density,
+                  subgroup_std = iota_subgroup_std)
+
+    return (truth, priors, node_effects, group_effects)
+
+def run_test(file_name, test_config, truth_in,
```
mbannick, line 93: All your test names need to start with `test_` otherwise `pytest` won't see them. So this should be changed to `test_run` or something. And then the above test that doesn't actually have any asserts in it should not start with the name `test_`.

To skip running this test when you're on travis, for example, you would have:
```
def test_run(file_name, test_config, truth_in, start_from_truth, test_asymptotic, dismod):
    ...
```
It knows to skip this test in travis because of the `dismod` argument.

gma-uw: create_database.py, dismod_tests.py, example_db.py are not tests. They are used by the other modules having names that start with test_   Is you comment still valid?

src/cascade_at/core/db.py
```diff
@@ -62,15 +69,24 @@ def __getattr__(self, name):
...
-swarm = ModuleProxy("jobmon.client.swarm")
-api = ModuleProxy("jobmon.client.api")
-task = ModuleProxy("jobmon.client.task")
-task_template = ModuleProxy("jobmon.client.task_template")
-sge = ModuleProxy("jobmon.client.execution.strategies.sge")
+
+import sys
+
+if 'darwin' in sys.platform:    # gma Something, perhaps db_tools, is importing db_queries incorrectly causing a deprecation error
+    import sys
```
zhengp0, line 81: Already import `sys` outside the `if`

src/cascade_at/core/db.py
```diff
@@ -62,15 +69,24 @@ def __getattr__(self, name):
...
+    del sys.modules['db_queries']
+    db_queries = ModuleProxy("db_queries")
+
+if 'darwin' in sys.platform:    # gma Add logic to skip jobmon imports if jobmon switch is not set at top level call
+    LOG.warning("FIXME -- gma -- Add logic to skip jobmon imports if jobmon switch is not set at top level call")
+else:
```
zhengp0, line 87: where is the `db_queries` go? seem like when is not macOS, you will not have `db_queries` anymore.

PR #316, opened 2020-11-18
Cv prior strategy fix

gma-uw:
I am not sure that this will result in the correct behavior, but at least it fixes 2 errors.


zhengp0 (approved):
Thanks @gma-uw ! Please merge these into the main branch. And I will try to deploy it and see if it fix the problem.

PR #317, opened 2020-11-25
MAINTAIN: add jobmon dependency with version 0.0.5

zhengp0:


mbannick (approved):
Thanks, @zhengp0 !

setup.py
```diff
@@ -16,15 +16,15 @@
         "intervaltree",
         "pytest",
         "tables",
-        "networkx"
+        "networkx",
+        "jobmon==0.0.5"
```
mbannick: this has to go in the `ihme` section below because travis isn't able to find it

mbannick (changes requested):

setup.py
```diff
@@ -26,7 +25,9 @@
             "db-queries",
             "gbd",
             "elmo",
-            "ihme-rules"],
+            "ihme-rules",
+            "jobmon==0.0.5"
```
zhengp0, line 29: Thanks @mbannick ! Let me know if this is good :)

mbannick: yes!! looks great!

mbannick (approved):

PR #318, opened 2020-11-30
When stats is an empty dataframe, the mulcov_stats.csv file has no co…

gma-uw:
…lumns which causes errors downstream. This occurs when a model has no country covariates.


zhengp0 (approved):
Thanks @gma-uw !

PR #319, opened 2020-12-01
Viz covariate as rate name mod

gma-uw:
This mod was required to handle a change to EpiVis covariate menus  made to show only rates (not integrands) in the measure menu when multiplier_type is rate_value. This messes up the code slightly as a rate name is not in a variable called "measure_id".


zhengp0 (approved):
Thanks @gma-uw !

PR #320, opened 2020-12-07
Missed converting a measure_id to an int.

gma-uw:
There was a second type conversion required for the rate menu change in EpiVis.


zhengp0 (approved):
Thanks @gma-uw !

PR #324, opened 2021-01-13
Fix one cov reference

gma-uw:


gma-uw:
I'm not sure why the omega changes are combined with this Peng.

PR #325, opened 2021-01-14
The reference value for the one covariate is supposed to be 0, otherw…

gma-uw:
…ise it has no effect.


gma-uw:
One line change is not worth reviewing

PR #328, opened 2021-01-16
Disable covariate max difference

gma-uw:
You may want to review this one Peng. Brad and I think this is more clear, though

### DIFF
--- a/src/cascade_at/inputs/measurement_inputs.py
+++ b/src/cascade_at/inputs/measurement_inputs.py
@@ -463,10 +463,8 @@ class MeasurementInputs:
                         covariate_dict={c.name: parent_df},
                         population_df=pop_df
                     )[c.name].iloc[0]
-                    max_difference = np.max(
-                        np.abs(all_loc_df.mean_value - reference_value)
-                    ) + CascadeConstants.PRECISION_FOR_REFERENCE_VALUES
-
+                    LOG.info(f"Setting covariate {c.name} max_difference = nan to disable data hold_out due to covariate value.")
+                    max_difference = np.nan
                 c.reference = reference_value
                 c.max_difference = max_difference
         covariate_specs.create_covariate_list()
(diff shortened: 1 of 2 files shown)


src/cascade_at/inputs/measurement_inputs.py
```diff
@@ -463,10 +463,8 @@ def calculate_country_covariate_reference_values(
                         covariate_dict={c.name: parent_df},
                         population_df=pop_df
                     )[c.name].iloc[0]
-                    max_difference = np.max(
-                        np.abs(all_loc_df.mean_value - reference_value)
-                    ) + CascadeConstants.PRECISION_FOR_REFERENCE_VALUES
-
+                    LOG.info(f"Setting covariate {c.name} max_difference = nan to disable data hold_out due to covariate value.")
+                    max_difference = np.nan
```
zhengp0, line 467: I think the original intention is to "outlier" data points by the value of the covariate being out of control. I can see that the weird covariate value might screw up the estimate of the covariate multiplier, but if this brings more troubles depends on `reference_value` and `PERCISION_FOR_REFERENCE_VALUES` I agree we could disable this.

zhengp0 (approved):

gma-uw:
Yes, but the problem was the code was setting the limits so it would include the full range of covariate values, assuming a reference value. So everything was included anyway. A null makes it clear all the covariates are included.

But as the reference value changes down the cascade, then that might cause some data to be thrown out by mistake.

So although the limit may have been intentionally set, the reasoning behind setting it was flawed.

On Jan 28, 2021, at 7:33 PM, Peng Zheng <notifications@github.com<mailto:notifications@github.com>> wrote:



@zhengp0 commented on this pull request.

________________________________

In src/cascade_at/inputs/measurement_inputs.py<https://github.com/ihmeuw/cascade-at/pull/328#discussion_r566557679>:

> @@ -463,10 +463,8 @@ def calculate_country_covariate_reference_values(
                         covariate_dict={c.name: parent_df},
                         population_df=pop_df
                     )[c.name].iloc[0]
-                    max_difference = np.max(
-                        np.abs(all_loc_df.mean_value - reference_value)
-                    ) + CascadeConstants.PRECISION_FOR_REFERENCE_VALUES
-
+                    LOG.info<http://LOG.info>(f"Setting covariate {c.name} max_difference = nan to disable data hold_out due to covariate value.")
+                    max_difference = np.nan


I think the original intention is to "outlier" data points by the value of the covariate being out of control. I can see that the weird covariate value might screw up the estimate of the covariate multiplier, but if this brings more troubles depends on reference_value and PERCISION_FOR_REFERENCE_VALUES I agree we could disable this.

PR #329, opened 2021-01-21
This error was preventing construction of the dismod.db, where these …

gma-uw:
…errors are more easily debugged.

### DIFF
--- a/src/cascade_at/inputs/uncertainty.py
+++ b/src/cascade_at/inputs/uncertainty.py
@@ -164,6 +164,6 @@ def stdev_from_crosswalk_version(crosswalk_version):
 
     # Do a final check on standard error
     if ((standard_error <= 0) | standard_error.isnull()).any():
-        raise ValueError("There are still negative / 0 / or null values for standard error.")
+        LOG.error("There are unrepairable negative or null values for standard error in the bundle.")
 
     return standard_error


src/cascade_at/inputs/uncertainty.py
```diff
@@ -164,6 +164,6 @@ def stdev_from_crosswalk_version(crosswalk_version):
 
     # Do a final check on standard error
     if ((standard_error <= 0) | standard_error.isnull()).any():
-        raise ValueError("There are still negative / 0 / or null values for standard error.")
+        LOG.error("There are unrepairable negative or null values for standard error in the bundle.")
```
zhengp0, line 167: Is `dismod.db` still useful in the presence of this error? I worried that this will screw up the downstream tasks. But I believe `dismod` does its own checking and if you think `dismod.db` is valuable this is alright :)

zhengp0 (approved):

gma-uw:
Yes, the db will have a bad std in it, but otherwise the modeler would have to go back to the bundle to find it. At least this way, if we have a database, we can fix the std and proceed.

On Jan 28, 2021, at 7:42 PM, Peng Zheng <notifications@github.com<mailto:notifications@github.com>> wrote:



@zhengp0 commented on this pull request.

________________________________

In src/cascade_at/inputs/uncertainty.py<https://github.com/ihmeuw/cascade-at/pull/329#discussion_r566560068>:

> @@ -164,6 +164,6 @@ def stdev_from_crosswalk_version(crosswalk_version):

     # Do a final check on standard error
     if ((standard_error <= 0) | standard_error.isnull()).any():
-        raise ValueError("There are still negative / 0 / or null values for standard error.")
+        LOG.error("There are unrepairable negative or null values for standard error in the bundle.")


Is dismod.db still useful in the presence of this error? I worried that this will screw up the downstream tasks. But I believe dismod does its own checking and if you think dismod.db is valuable this is alright :)

